PR #182, opened 2020-08-05
TOSS-common

liuyu85cn:
in progress, for interface / code structure review only. 

Add 2 new interface in thrift 
    
ExecResponse addEdgesAtomic(1: AddEdgesRequest req);
ExecResponse processTransaction(1: TransactionReq txnCtx);

addEdgesAtomic is for client outside storage cluster. 
processTransaction is used internal in storage cluster.

### DIFF
--- a/src/common/clients/storage/GraphStorageClient.cpp
+++ b/src/common/clients/storage/GraphStorageClient.cpp
@@ -34,11 +34,13 @@ GraphStorageClient::getNeighbors(GraphSpaceID space,
         });
 
     if (!status.ok()) {
+        LOG(INFO) << "messi !status.ok()";
         return folly::makeFuture<StorageRpcResponse<cpp2::GetNeighborsResponse>>(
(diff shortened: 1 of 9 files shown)


src/common/clients/storage/GraphStorageClient.cpp
```diff
@@ -170,6 +173,63 @@ GraphStorageClient::addEdges(GraphSpaceID space,
...
+            return client->future_addEdgesAtomic(r);
+        },
+        [] (const std::pair<const PartitionID, std::vector<cpp2::NewEdge>>& p) {
+            return p.first;
+        });
+}
+
+
+folly::Future<StatusOr<storage::cpp2::ExecResponse>>
+GraphStorageClient::forwardTransaction(cpp2::TransactionReq txnCtx,
```
dangleptr, line 218: I suggest we move the storage internal interfaces definition into nebula-storage.

PR #196, opened 2020-08-17
Delete hdfs directory

laura-ding:


darionyaphet:
If so how to download SST files which will be ingest into nebula?

laura-ding:
@darionyaphet http://jira.nebula-graph.io:8081/projects/NG/issues/NG-98?filter=allissues

laura-ding:
2.0 is also need, so close it.

PR #202, opened 2020-08-19
Support integer vertex id.

CPWstatic:
Close #190

### DIFF
--- a/src/common/clients/meta/MetaClient.cpp
+++ b/src/common/clients/meta/MetaClient.cpp
@@ -1664,6 +1664,25 @@ StatusOr<int32_t> MetaClient::getSpaceVidLen(const GraphSpaceID& spaceId) {
     return vIdLen;
 }
 
+StatusOr<Value::Type> MetaClient::getSpaceVidType(const GraphSpaceID& spaceId) {
+    if (!ready_) {
+        return Status::Error("Not ready!");
+    }
+    folly::RWSpinLock::ReadHolder holder(localCacheLock_);
+    auto spaceIt = localCache_.find(spaceId);
+    if (spaceIt == localCache_.end()) {
+        LOG(ERROR) << "Space " << spaceId << " not found!";
+        return Status::Error(folly::stringPrintf("Space %d not found", spaceId));
+    }
+    auto vIdType = spaceIt->second->vertexIdType_;
+    if (vIdType != Value::Type::INT || vIdType != Value::Type::STRING) {
+        std::stringstream ss;
+        ss << "Space " << spaceId << "vertexId type invalid: " << vIdType;
+        return Status::Error(ss.str());
+    }
+    return vIdType;
+}
+
 StatusOr<std::shared_ptr<const NebulaSchemaProvider>>
 MetaClient::getTagSchemaFromCache(GraphSpaceID spaceId, TagID tagID, SchemaVer ver) {
     if (!ready_) {
--- a/src/common/clients/meta/MetaClient.cpp
+++ b/src/common/clients/meta/MetaClient.cpp
@@ -231,24 +231,8 @@ bool MetaClient::loadData() {
             LOG(ERROR) << "Get space properties failed for space " << spaceId;
             return false;
         }
-        const auto& properties = resp.value().get_properties();
-        SpaceDesc spaceDesc;
-        spaceDesc.spaceName_ = properties.get_space_name();
-        spaceDesc.partNum_ = properties.get_partition_num();
-        spaceDesc.replicaFactor_ = properties.get_replica_factor();
-        spaceDesc.charsetName_ = properties.get_charset_name();
-        spaceDesc.collationName_ = properties.get_collate_name();
-        spaceDesc.vidSize_ = properties.get_vid_size();
-        if (properties.get_vid_type() == cpp2::PropertyType::STRING) {
-            spaceDesc.vidType_ = Value::Type::STRING;
-        } else if (properties.get_vid_type() == cpp2::PropertyType::INT64) {
-            spaceDesc.vidType_ = Value::Type::INT;
-        } else {
-            LOG(ERROR) << "Unsupported vid type: "
-                << meta::cpp2::_PropertyType_VALUES_TO_NAMES.at(properties.get_vid_type());
-            return false;
-        }
-        spaceCache->spaceDesc_ = std::move(spaceDesc);
+        auto properties = resp.value().get_properties();
+        spaceCache->spaceDesc_ = std::move(properties);
 
         cache.emplace(spaceId, spaceCache);
         spaceIndexByName.emplace(space.second, spaceId);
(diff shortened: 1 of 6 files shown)


critical27:
I see you replace VertexID with Value. Is this different from we discussed before? I thought we use VertexID, use Integer Vid as string. And storage will return Value with correct vid type (int or string).

CPWstatic:
> I see you replace VertexID with Value. Is this different from we discussed before? I thought we use VertexID, use Integer Vid as string. And storage will return Value with correct vid type (int or string).
> 
> PS: the perf between int id and string id is not trivial according to benchmark.

The storage client will convert all int vid to string vid, there are not much change.

src/common/clients/meta/MetaClient.cpp
```diff
@@ -200,7 +200,7 @@ bool MetaClient::loadData() {
 
         auto spaceCache = std::make_shared<SpaceInfoCache>();
         auto partsAlloc = r.value();
-        spaceCache->spaceName = space.second;
+        auto spaceName = space.second;
```
laura-ding, line 203: auto&

src/common/clients/storage/GraphStorageClient.cpp
```diff
@@ -515,5 +529,148 @@ GraphStorageClient::lookupAndTraverse(GraphSpaceID space,
...
+
+StatusOr<std::function<const VertexID&(const cpp2::NewVertex&)>>
+GraphStorageClient::getIdFromNewVertex(GraphSpaceID space) const {
+    auto vidTypeStatus = metaClient_->getSpaceVidType(space);
+    if (!vidTypeStatus) {
+        return vidTypeStatus.status();
+    }
+    auto vidType = std::move(vidTypeStatus).value();
+
+    std::function<const VertexID&(const cpp2::NewVertex&)> cb;
```
laura-ding, line 570: `const cpp2::NewVertex&` the const look no necessary because the cb maybe modify the value, It's easier to understand without const.

src/common/clients/storage/GraphStorageClient.cpp
```diff
@@ -515,5 +529,148 @@ GraphStorageClient::lookupAndTraverse(GraphSpaceID space,
...
+                return mutableEK.src.getStr();
+            };
+    } else if (vidType == meta::cpp2::PropertyType::FIXED_STRING) {
+        cb = [](const cpp2::EdgeKey& eKey) -> const VertexID& {
+                DCHECK_EQ(Value::Type::STRING, eKey.src.type());
+                DCHECK_EQ(Value::Type::STRING, eKey.dst.type());
+                return eKey.src.getStr();
+            };
+    } else {
+        return Status::Error("Only support integer/string type vid.");
```
laura-ding, line 644: Maybe never reach here

laura-ding (approved):

src/common/clients/storage/GraphStorageClient.cpp
```diff
@@ -515,5 +529,148 @@ GraphStorageClient::lookupAndTraverse(GraphSpaceID space,
...
+    }
+    auto vidType = std::move(vidTypeStatus).value();
+
+    std::function<const VertexID&(const Row&)> cb;
+    if (vidType == meta::cpp2::PropertyType::INT64) {
+        cb = [](const Row& r) -> const VertexID& {
+                // The first column has to be the vid
+                DCHECK_EQ(Value::Type::INT, r.values[0].type());
+                auto& mutableR = const_cast<Row&>(r);
+                mutableR.values[0] = Value(reinterpret_cast<const char*>(&r.values[0].getInt()));
```
dutor, line 546: What if the integer has zero bytes?

CPWstatic: In which situation?

dutor (approved):
Awesome

src/common/clients/meta/MetaClient.cpp
```diff
@@ -532,8 +532,8 @@ void MetaClient::getResponse(Request req,
                 } else {
                     LOG(ERROR) << "Send request to " << host << ", exceed retry limit";
                     pro.setValue(
-                        Status::Error(folly::stringPrintf("RPC failure in MetaClient: %s",
-                                                          t.exception().what().c_str())));
+                        Status::Error("RPC failure in MetaClient: %s",
```
laura-ding, line 535: merge the rows

src/common/clients/meta/MetaClient.cpp
```diff
@@ -920,9 +910,8 @@ StatusOr<TagID> MetaClient::getTagIDByNameFromCache(const GraphSpaceID& space,
     folly::RWSpinLock::ReadHolder holder(localCacheLock_);
     auto it = spaceTagIndexByName_.find(std::make_pair(space, name));
     if (it == spaceTagIndexByName_.end()) {
-        std::string error = folly::stringPrintf("TagName `%s'  is nonexistent",
+        return Status::Error("TagName `%s'  is nonexistent",
```
laura-ding, line 913: ditto

src/common/clients/meta/MetaClient.cpp
```diff
@@ -951,9 +939,8 @@ StatusOr<EdgeType> MetaClient::getEdgeTypeByNameFromCache(const GraphSpaceID& sp
     folly::RWSpinLock::ReadHolder holder(localCacheLock_);
     auto it = spaceEdgeIndexByName_.find(std::make_pair(space, name));
     if (it == spaceEdgeIndexByName_.end()) {
-        std::string error = folly::stringPrintf("EdgeName `%s'  is nonexistent",
+        return Status::Error("EdgeName `%s'  is nonexistent",
```
laura-ding, line 942: ditto

src/common/interface/meta.thrift
```diff
@@ -163,18 +163,19 @@ struct IdName {
...
-    2: i32                  partition_num,
-    3: i32                  replica_factor,
-    4: i32                  vid_size = 8,
-    5: binary               charset_name,
-    6: binary               collate_name,
+    2: i32                  partition_num = 0,
+    3: i32                  replica_factor = 0,
+    4: binary               charset_name,
+    5: binary               collate_name,
+    6: i32                  vid_size = 8,
```
laura-ding, line 172: need to modify to i16

laura-ding (approved):

src/common/interface/meta.thrift
```diff
@@ -163,18 +163,19 @@ struct IdName {
...
+struct SpaceDesc {
     1: binary               space_name,
-    2: i32                  partition_num,
-    3: i32                  replica_factor,
-    4: i32                  vid_size = 8,
-    5: binary               charset_name,
-    6: binary               collate_name,
+    2: i32                  partition_num = 0,
+    3: i32                  replica_factor = 0,
+    4: binary               charset_name,
```
darionyaphet, line 170: AFAIK modify the order of element in thrift structure maybe not a good idea.

src/common/clients/storage/GraphStorageClient.h
```diff
@@ -123,6 +123,22 @@ class GraphStorageClient : public StorageClientBase<cpp2::GraphStorageServiceAsy
         cpp2::IndexSpec indexSpec,
         cpp2::TraverseSpec traverseSpec,
         folly::EventBase* evb = nullptr);
+
+private:
+    StatusOr<std::function<const VertexID&(const Row&)>>
```
darionyaphet, line 128: alignment ?

PR #206, opened 2020-08-21
Fix error code.

CPWstatic:


src/common/interface/graph.thrift
```diff
@@ -41,12 +41,12 @@ enum ErrorCode {
     // Nothing is executed When command is comment
     E_STATEMENT_EMTPY = -9,
 
-    // User and permission error
-    E_USER_NOT_FOUND = -10,
-    E_BAD_PERMISSION = -11,
-
     // semantic error
-    E_SEMANTIC_ERROR = -12,
+    E_SEMANTIC_ERROR = -10,
```
Shylock-Hg, line 45: If so, make it different with 1.0

PR #207, opened 2020-08-24
Add AttributeExpression

dutor:
* Add `AttributeExpression` to denote the form `<expr>.label`
 * Add `EdgeExpression` and `VertexExpression` to represent the Vertex and Edge object internally.
 * Rename `SymbolPropertyExpression` to `PropertyExpression`, AKA. the base class of all property-related expressions.
 * Add `LabelAttributeExpression` to denote an unresolved expression of the form `label.label`. Thus all `PropertyExpression` are deterministic now.

### DIFF
--- a/src/common/expression/LabelAttributeExpression.cpp
+++ b/src/common/expression/LabelAttributeExpression.cpp
@@ -9,8 +9,6 @@
 namespace nebula {
 
 std::string LabelAttributeExpression::toString() const {
-    CHECK(left()->kind() == Kind::kLabel);
-    CHECK(right()->kind() == Kind::kLabel);
     return left()->toString() + "." + right()->toString();
 }
 
--- a/src/common/expression/LabelAttributeExpression.cpp
+++ b/src/common/expression/LabelAttributeExpression.cpp
@@ -1,0 +1,17 @@
+/* Copyright (c) 2020 vesoft inc. All rights reserved.
+ *
+ * This source code is licensed under Apache 2.0 License,
+ * attached with Common Clause Condition 1.0, found in the LICENSES directory.
+ */
+
+#include "common/expression/LabelAttributeExpression.h"
+
+namespace nebula {
+
+std::string LabelAttributeExpression::toString() const {
+    CHECK(left()->kind() == Kind::kLabel || left()->kind() == Kind::kConstant);
+    CHECK(right()->kind() == Kind::kLabel || right()->kind() == Kind::kConstant);
+    return left()->toString() + "." + right()->toString();
+}
+
+}   // namespace nebula
--- a/src/common/expression/LabelAttributeExpression.cpp
+++ b/src/common/expression/LabelAttributeExpression.cpp
@@ -9,8 +9,8 @@
 namespace nebula {
 
 std::string LabelAttributeExpression::toString() const {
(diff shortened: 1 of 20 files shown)


src/common/expression/LabelAttributeExpression.cpp
```diff
@@ -0,0 +1,17 @@
...
+ * attached with Common Clause Condition 1.0, found in the LICENSES directory.
+ */
+
+#include "common/expression/LabelAttributeExpression.h"
+
+namespace nebula {
+
+std::string LabelAttributeExpression::toString() const {
+    CHECK(left()->kind() == Kind::kLabel || left()->kind() == Kind::kConstant);
+    CHECK(right()->kind() == Kind::kLabel || right()->kind() == Kind::kConstant);
```
CPWstatic, line 13: It seems not right if the left/right are `constant`, according to `leftLabel ` and `rightLabel `.

src/common/expression/LabelAttributeExpression.h
```diff
@@ -0,0 +1,39 @@
...
+#include "common/expression/BinaryExpression.h"
+#include "common/expression/LabelExpression.h"
+
+namespace nebula {
+
+// label.label
+class LabelAttributeExpression final : public BinaryExpression {
+public:
+    explicit LabelAttributeExpression(Expression *lhs = nullptr,
+                                      Expression *rhs = nullptr)
```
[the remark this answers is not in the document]
dutor: > Would it be better make the ctor only accept `LabelExpression `.

Yes, we could. But since it derives from `BinaryExpression`, `left()` and `right()` are `Expression*`. So I think we hardly benefit from doing so.

src/common/expression/LabelAttributeExpression.cpp
Shylock-Hg: Agree, and better check in ctor.

dutor: > It seems not right if the left/right are `constant`, according to `leftLabel ` and `rightLabel `.

Yeah, you are right. I thought we might construct it from two constant expression in UT like others. But it semms not necessary. I will address this.

src/common/expression/LabelAttributeExpression.h
```diff
@@ -0,0 +1,39 @@
...
+
+#ifndef COMMON_EXPRESSION_LABELATTRIBUTEEXPRESSION_H_
+#define COMMON_EXPRESSION_LABELATTRIBUTEEXPRESSION_H_
+
+#include "common/expression/BinaryExpression.h"
+#include "common/expression/LabelExpression.h"
+
+namespace nebula {
+
+// label.label
```
Shylock-Hg, line 15: This seems had be contained in AttributeExpression, why need a separate class?

src/common/expression/LabelAttributeExpression.h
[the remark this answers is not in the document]
dutor: Make it derive from Expression.

src/common/expression/LabelAttributeExpression.h
dutor: In order to find out the expressions that need to be rewritten.

CPWstatic (approved):

yixinglu (approved):
LGTM

PR #209, opened 2020-08-26
Fix operator name

dutor:
as title.

### DIFF
--- a/src/common/expression/LogicalExpression.cpp
+++ b/src/common/expression/LogicalExpression.cpp
@@ -37,7 +37,7 @@ std::string LogicalExpression::toString() const {
             op = "||";
             break;
         case Kind::kLogicalXor:
-            op = "XOR";
+            op = " XOR ";
             break;
         default:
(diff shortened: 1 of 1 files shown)


darionyaphet (approved):

src/common/expression/LogicalExpression.cpp
```diff
@@ -37,7 +37,7 @@ std::string LogicalExpression::toString() const {
             op = "||";
             break;
         case Kind::kLogicalXor:
-            op = "XOR";
+            op = " XOR ";
```
darionyaphet, line 40: another operation name？

yixinglu (approved):

src/common/expression/LogicalExpression.cpp
dutor: > another operation name？

What do you mean another?

Previously, there are two format to denote the logical xor operator, i.e. `^` and `XOR`. But now `^` are removed since it conflicts with the semantic of Cypher

PR #215, opened 2020-09-02
Add clone interface for expression

yixinglu:
This clone is a deep copy of original expression object.

To simplify the following use case:

```cpp
Expression *expr = ...;
auto cloneExpr = Expression::decode(Expression::encode(*expr));
// =>
auto cloneExpr = expr->clone();
```


Shylock-Hg:
The ExpressionUtils had this already.

dutor (approved):

yixinglu:
> The ExpressionUtils had this already.

still not too simple to use.

laura-ding (approved):

PR #217, opened 2020-09-04
Add function size.

CPWstatic:


src/common/function/test/FunctionManagerTest.cpp
```diff
@@ -144,6 +148,52 @@ TEST_F(FunctionManagerTest, functionCall) {
         auto res = std::move(result).value()({false});
         EXPECT_EQ(res, 0);
     }
+    {
```
Shylock-Hg, line 151: Add cases about invalid data type.

laura-ding (approved):

yixinglu (approved):

laura-ding (approved):

dutor (approved):
Awesome

PR #218, opened 2020-09-07
Accept multiple visitors in once traversal.

Shylock-Hg:


yixinglu:
I don't know why we need this accept function?  If you want to do so, why not do it outside?

Shylock-Hg:
Emm.Still multiple traversal so close it.

PR #221, opened 2020-09-09
Add methods for path type.

CPWstatic:


src/common/datatypes/Path.cpp
```diff
@@ -7,5 +7,28 @@
...
+void Path::reverse() {
+    if (steps.empty()) {
+        return;
+    }
+    std::reverse(steps.begin(), steps.end());
+    swap(src, steps.front().dst);
+    for (size_t i = 0; i < steps.size(); ++i) {
+        if ((i + 2) <= steps.size()) {
+            swap(steps[i].dst, steps[i + 1].dst);
+        }
```
dutor, line 19: `if ( i + 1 < steps.size())` is more clear and consistent with the loop condition and swap.

dutor: BTW. If the condition is `i < steps.size() - 1`, I think you could remove the `if`, and reverse the last edge after the loop.

src/common/datatypes/Path.cpp
```diff
@@ -7,5 +7,28 @@
...
+        steps[i].type = -steps[i].type;
+    }
+}
 
+bool Path::append(Path path) {
+    if (src != path.src && steps.back().dst != path.src) {
+        return false;
+    }
```
dutor, line 27: Better add some comments about the semantics of this interface.

dutor (dismissed):
LGTM.

src/common/datatypes/Path.cpp
```diff
@@ -7,5 +7,27 @@
...
+        steps[i].type = -steps[i].type;
+    }
+    steps.back().type = -steps.back().type;
+}
 
+bool Path::append(Path path) {
+    if (src != path.src && steps.back().dst != path.src) {
+        return false;
+    }
+    steps.reserve(path.steps.size());
```
Shylock-Hg, line 27: steps.size() + path.steps.size()

CPWstatic: The `reserve` is about to increase the capacity.

Shylock-Hg: The parameter is the new capacity instead of increase capacity.

laura-ding (approved):

src/common/datatypes/Path.cpp
yixinglu: steps.size() + path.steps.size() - 1

Shylock-Hg (approved):

dutor (approved):

PR #227, opened 2020-09-14
Add some methods for edge type.

CPWstatic:


src/common/datatypes/Edge.h
```diff
@@ -85,6 +85,22 @@ struct Edge {
                ranking == rhs.ranking &&
                props == rhs.props;
     }
+
+    void format() {
+        if (type < 0) {
+            type = -type;
```
Shylock-Hg, line 91: You can call reverse here directly.

laura-ding, line 91: use call `reverse()` directly

nevermore3 (approved):

dutor (approved):

PR #228, opened 2020-09-14
statistics index in common

panda-sheep:
#NG-167

### DIFF
--- a/src/common/clients/meta/MetaClient.cpp
+++ b/src/common/clients/meta/MetaClient.cpp
@@ -421,12 +421,12 @@ bool MetaClient::loadIndexes(GraphSpaceID spaceId,
 }
 
 
-Status MetaClient::checkTagIndexed(GraphSpaceID space, TagID tagID) {
+Status MetaClient::checkTagIndexed(GraphSpaceID spaceId, IndexID indexID) {
     folly::RWSpinLock::ReadHolder holder(localCacheLock_);
-    auto it = localCache_.find(space);
+    auto it = localCache_.find(spaceId);
     if (it != localCache_.end()) {
-        auto tagIt = it->second->tagIndexes_.find(tagID);
-        if (tagIt != it->second->tagIndexes_.end()) {
+        auto indexIt = it->second->tagIndexes_.find(indexID);
+        if (indexIt != it->second->tagIndexes_.end()) {
             return Status::OK();
         } else {
             return Status::IndexNotFound();
@@ -436,12 +436,12 @@ Status MetaClient::checkTagIndexed(GraphSpaceID space, TagID tagID) {
 }
 
 
-Status MetaClient::checkEdgeIndexed(GraphSpaceID space, EdgeType edgeType) {
+Status MetaClient::checkEdgeIndexed(GraphSpaceID spaceId, IndexID indexID) {
     folly::RWSpinLock::ReadHolder holder(localCacheLock_);
-    auto it = localCache_.find(space);
+    auto it = localCache_.find(spaceId);
     if (it != localCache_.end()) {
-        auto edgeIt = it->second->edgeIndexes_.find(edgeType);
-        if (edgeIt != it->second->edgeIndexes_.end()) {
+        auto indexIt = it->second->edgeIndexes_.find(indexID);
+        if (indexIt != it->second->edgeIndexes_.end()) {
             return Status::OK();
         } else {
             return Status::IndexNotFound();
@@ -1414,12 +1414,14 @@ MetaClient::dropEdgeSchema(GraphSpaceID spaceId, std::string name, const bool if
 
 folly::Future<StatusOr<IndexID>>
 MetaClient::createTagIndex(GraphSpaceID spaceID,
+                           cpp2::IndexType indexType,
                            std::string indexName,
                            std::string tagName,
                            std::vector<std::string> fields,
                            bool ifNotExists) {
     cpp2::CreateTagIndexReq req;
     req.set_space_id(spaceID);
+    req.set_index_type(std::move(indexType));
     req.set_index_name(std::move(indexName));
     req.set_tag_name(std::move(tagName));
     req.set_fields(std::move(fields));
@@ -1544,12 +1546,14 @@ MetaClient::listTagIndexStatus(GraphSpaceID spaceID) {
 
 folly::Future<StatusOr<IndexID>>
 MetaClient::createEdgeIndex(GraphSpaceID spaceID,
+                            cpp2::IndexType indexType,
                             std::string indexName,
                             std::string edgeName,
                             std::vector<std::string> fields,
                             bool ifNotExists) {
     cpp2::CreateEdgeIndexReq req;
     req.set_space_id(spaceID);
+    req.set_index_type(std::move(indexType));
     req.set_index_name(std::move(indexName));
     req.set_edge_name(std::move(edgeName));
     req.set_fields(std::move(fields));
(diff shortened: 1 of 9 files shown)


src/common/interface/meta.thrift
```diff
@@ -570,79 +571,96 @@ struct HBReq {
     5: binary     git_info_sha
 }
 
+enum IndexType {
+    NORMAL       = 0x01,
+    VERTEXCOUNT  = 0x02,
+    EDGECOUNT    = 0x03,
```
darionyaphet, line 577: `EDGE_COUNT` would be better ?

panda-sheep: Good point，👍

src/common/meta/Common.h
```diff
@@ -29,6 +29,13 @@ struct PartHosts {
     }
 };
 
+enum class IndexType : int8_t {
```
darionyaphet, line 32: It's same with thrift's source code. Would you using only once ?

src/common/interface/storage.thrift
```diff
@@ -604,9 +604,36 @@ struct LookupAndTraverseRequest {
...
+struct GetEdgesStatisRequest {
+    1: required common.GraphSpaceID         space_id,
+    2: required list<common.PartitionID>    parts,
+    // Return the statistical value of one index_id
+    3: required common.IndexID              index,
+}
+
+struct GetStatisResponse {
+    1: required ResponseCommon              result,
+    2: required i64                         count,
```
darionyaphet, line 630: Because maybe support count sum man min or any other something. `result` maybe a better name？

src/common/clients/meta/MetaClient.cpp
```diff
@@ -56,7 +56,7 @@ MetaClient::~MetaClient() {
 
 bool MetaClient::isMetadReady() {
     auto ret = heartbeat().get();
-    if (!ret.ok() && ret.status() != Status::LeaderChanged()) {
+    if (!ret.ok() && ret.status() !=   Status::LeaderChanged()) {
```
darionyaphet, line 59: extra space

src/common/clients/meta/MetaClient.cpp
```diff
@@ -1462,9 +1542,12 @@ MetaClient::dropTagIndex(GraphSpaceID spaceID, std::string name, bool ifExists)
 
 
 folly::Future<StatusOr<cpp2::IndexItem>>
-MetaClient::getTagIndex(GraphSpaceID spaceID, std::string name) {
+MetaClient::getTagIndex(GraphSpaceID spaceId,
+                        cpp2::IndexType indexType,
+                        std::string name) {
     cpp2::GetTagIndexReq req;
-    req.set_space_id(std::move(spaceID));
+    req.set_space_id(std::move(spaceId));
```
darionyaphet, line 1549: why move an integer ? any benefit ?

src/common/interface/meta.thrift
```diff
@@ -571,79 +572,96 @@ struct HBReq {
...
```
darionyaphet, line 578: Do we need to distinguish the edge and vertex's count ？

src/common/interface/meta.thrift
```diff
@@ -571,79 +572,96 @@ struct HBReq {
...
+    GROUPBYCOUNT = 0x04, // out/in-degree
+}
+
 struct CreateTagIndexReq {
     1: common.GraphSpaceID  space_id,
-    2: binary               index_name,
-    3: binary               tag_name,
-    4: list<binary>			fields,
-    5: bool                 if_not_exists,
+    2: IndexType            index_type,
```
darionyaphet, line 584: The `Statistics Index` will created by user ?

src/common/meta/IndexManager.h
```diff
@@ -13,8 +13,17 @@
 
 namespace nebula {
 namespace meta {
+
 using IndexItem = cpp2::IndexItem;
 
+enum class IndexType : int8_t {
```
laura-ding, line 19: I think it's an unnecessary definition

src/common/meta/IndexManager.h
```diff
@@ -30,20 +39,22 @@ class IndexManager {
...
 
     virtual StatusOr<std::vector<std::shared_ptr<IndexItem>>>
-    getTagIndexes(GraphSpaceID space) = 0;
+    getTagIndexes(GraphSpaceID space, cpp2::IndexType indexType) = 0;
 
     virtual StatusOr<std::vector<std::shared_ptr<IndexItem>>>
-    getEdgeIndexes(GraphSpaceID space) = 0;
+    getEdgeIndexes(GraphSpaceID space, cpp2::IndexType indexType) = 0;
 
+    /*
```
laura-ding, line 47: Why comment it out

src/common/interface/meta.thrift
panda-sheep: Automatically created

src/common/interface/storage.thrift
panda-sheep: Good point

src/common/meta/Common.h
panda-sheep: Here is used to specify the size as one byte

src/common/clients/meta/MetaClient.cpp
panda-sheep: Good catch

src/common/clients/meta/MetaClient.cpp
panda-sheep: I did not add it here, but I will remove `move`

src/common/meta/IndexManager.h
panda-sheep: Good point，fixed

src/common/interface/meta.thrift
```diff
@@ -607,15 +617,16 @@ struct ListTagIndexesReq {
...
 }
 
 struct CreateEdgeIndexReq {
-    1: common.GraphSpaceID 	space_id,
-    2: binary              	index_name,
-    3: binary              	edge_name,
-    4: list<binary>			fields,
-    5: bool                	if_not_exists,
+    1: common.GraphSpaceID  space_id,
+    2: IndexType            index_type,
```
bright-starry-sky, line 625: A question，Why to add `index_type` at here? I think statistic indexes are created automatically when creating Space or tag. right ? If so, I think a separate interface is better. right?

panda-sheep: Good question

> I think statistic indexes are created automatically when creating Space or tag. right ?
yeah

> If so, I think a separate interface is better. right?
In this way, a set of code logic is used, and the implementation is simpler and clearer

src/common/meta/IndexManager.h
panda-sheep: I did not find a place to call the function in the code

src/common/interface/meta.thrift
```diff
@@ -574,12 +575,22 @@ struct HBReq {
     5: binary     git_info_sha
 }
 
+enum IndexType {
+    ALL            = 0x01, // all tag indextype or all edge indextype
+    NORMAL         = 0x02,
+    VERTEX         = 0x03,
+    EDGE           = 0x04,
+    VERTEX_COUNT   = 0x05,
+    EDGE_COUNT     = 0x06,
```
critical27, line 584: Would you add some comment about each edge type?

panda-sheep: I will add comments for it.

src/common/interface/storage.thrift
panda-sheep: line 629 has existed  result

panda-sheep:
Reimplement, close this pr.

PR #234, opened 2020-09-16
containerExpression MoveItems

nevermore3:


yixinglu (commented):
Why need to move the items to outside ?

nevermore3:
> Why need to move the items to outside ?


Will traverse items and reset items in goValidator

yixinglu:
> > Why need to move the items to outside ?
> 
> Will traverse items and reset items in goValidator

I don't think it's a good practice to move some fields of these objects. you can define some get functions which must by called by moving object itself like following:

```cpp
std::vector<std::unique_ptr<Expression>> get() && {
  return std::move(items_);
}
```

yixinglu (approved):

dutor (approved):

PR #237, opened 2020-09-17
Support getSchemaIDByName in SchemaManager

darionyaphet:
This interface will be used by `Graph Layer` and `Storage Layer` later. 

It's to avoid some duplicate check when operate the indexes.

### DIFF
--- a/src/common/meta/SchemaManager.cpp
+++ b/src/common/meta/SchemaManager.cpp
@@ -16,5 +16,19 @@ std::unique_ptr<SchemaManager> SchemaManager::create(MetaClient *client) {
     return mgr;
 }
 
+StatusOr<std::pair<bool, int32_t>>
+SchemaManager::getSchemaIDByName(GraphSpaceID space, folly::StringPiece schemaName) {
+    auto ret = toEdgeType(space, schemaName);
+    if (ret.ok()) {
+        return std::make_pair(true, ret.value());
+    } else {
+        ret = toTagID(space, schemaName);
+        if (ret.ok()) {
+            return std::make_pair(false, ret.value());
+        }
+    }
+    return Status::Error("Schema not exist: %s", schemaName.str().c_str());
+}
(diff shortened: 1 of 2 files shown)


src/common/meta/SchemaManager.cpp
```diff
@@ -16,5 +16,19 @@ std::unique_ptr<SchemaManager> SchemaManager::create(MetaClient *client) {
     return mgr;
 }
 
+StatusOr<std::pair<bool, int32_t>>
```
bright-starry-sky, line 19: Could you help add comments for bool? Thanks.

bright-starry-sky:
Nice job. This method will make it easy to check what type of schema. Used for index-related operations.

src/common/meta/SchemaManager.cpp
darionyaphet: Sure ~

bright-starry-sky (approved):
LGTM

src/common/meta/SchemaManager.h
```diff
@@ -61,6 +61,11 @@ class SchemaManager {
     // get all version of all edge schema
     virtual StatusOr<EdgeSchemas> getAllVerEdgeSchema(GraphSpaceID space) = 0;
 
+    // Get the TagID or EdgeType by the name.
+    // The first one is a bool which is used to distinguish the type.
+    // When the result is an edge, it's true, otherwise it's false.
+    StatusOr<std::pair<bool, int32_t>>
+    getSchemaIDByName(GraphSpaceID space, folly::StringPiece schemaName);
```
dutor, line 68: In what kind of situation this method is used? If I were correct, a tag is allowed to have the same name with an edge.

dutor: I am wrong

dutor (approved):
LGTM.

PR #242, opened 2020-09-21
Operator< for vertex/edge/path

CPWstatic:


yixinglu (approved):

nevermore3 (dismissed):

src/common/datatypes/Edge.h
```diff
@@ -98,6 +98,11 @@ struct Edge {
         src = std::move(dst);
         dst = std::move(tmp);
     }
+
+    bool operator<(const Edge& rhs) const {
+        return src < rhs.src && dst < rhs.dst && type < rhs.type && ranking < rhs.ranking &&
+               props.size() < rhs.props.size();
+    }
```
dutor, line 105: I don't think this is the right way to implement less for a multi-field class. We should compare the next field only if the previous one is equal.

dutor (approved):
LGTM.

nevermore3 (approved):

yixinglu (approved):

PR #245, opened 2020-09-22
Support zone

darionyaphet:


src/common/interface/meta.thrift
```diff
@@ -826,6 +828,85 @@ struct ListIndexStatusResp {
...
+    1: binary                 group_name,
+}
+
+struct ListZonesResp {
+    1: ErrorCode        code,
+    2: common.HostAddr  leader,
+    3: list<Zone>       zones,
+}
+
+struct DrainZoneReq {
```
critical27, line 873: According to design doc, DrainZoneReq should only used in default mode?

src/common/interface/meta.thrift
```diff
@@ -826,6 +828,85 @@ struct ListIndexStatusResp {
...
+    1: binary        group_name,
+    2: binary        zone_name,
+}
+
+struct AddGroupReq {
```
critical27, line 878: According to design doc, should we specify which group is mirror?

src/common/interface/meta.thrift
```diff
@@ -336,23 +337,24 @@ struct AdminJobResp {
...
 struct ListSpacesReq {
 }
 
 struct ListSpacesResp {
-    1: ErrorCode code,
+    1: ErrorCode        code,
     // Valid if ret equals E_LEADER_CHANGED.
     2: common.HostAddr  leader,
-    3: list<IdName> spaces,
+    3: list<IdName>     spaces,
```
liuyu85cn, line 357: I'm not sure if we need to show zone info also.

src/common/interface/meta.thrift
darionyaphet: Currently We can support replication number is greater or equal with zone number. So it's ok

src/common/interface/meta.thrift
darionyaphet: I will support it in async copy module

src/common/interface/meta.thrift
darionyaphet: Sure I will support it later

critical27 (approved):
LGTM

liuyu85cn (approved):
LGTM

PR #246, opened 2020-09-22
scan interface

critical27:
As title, the interface is almost same as 1.0.

### DIFF
--- a/src/common/interface/storage.thrift
+++ b/src/common/interface/storage.thrift
@@ -607,6 +607,66 @@ struct LookupAndTraverseRequest {
  * End of Index section
  */
 
+struct ScanVertexRequest {
+    1: common.GraphSpaceID                  space_id,
+    2: common.PartitionID                   part_id,
+    // start key of this block
+    3: optional binary                      cursor,
+    4: list<VertexProp>                     return_columns,
+    // If no_columns is true, no properties of tags will be returned,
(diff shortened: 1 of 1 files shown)


src/common/interface/storage.thrift
```diff
@@ -607,6 +607,66 @@ struct LookupAndTraverseRequest {
  * End of Index section
  */
 
+struct ScanVertexRequest {
+    1: common.GraphSpaceID                  space_id,
+    2: common.PartitionID                   part_id,
+    // start key of this block
+    3: optional binary                      cursor,
+    4: list<VertexProp>                     return_columns,
```
darionyaphet, line 615: If so, you can support multi tag when running scan once?

bright-starry-sky (approved):
LGTM.

src/common/interface/storage.thrift
critical27: Yeah, so does 1.0

darionyaphet: Maybe only support one tag with multi column is enough. 

For Spark someone fetch multi tags into one frame.  IMO, this interface maybe make the user confuse.

critical27: Yeah, we could just add only one tag in the list in that case.

darionyaphet (approved):

PR #247, opened 2020-09-23
Set default value for vid type in space description

yixinglu:
fix error in vesoft-inc/nebula-storage#147

### DIFF
--- a/src/common/interface/meta.thrift
+++ b/src/common/interface/meta.thrift
@@ -173,7 +173,7 @@ struct SpaceDesc {
     3: i32                  replica_factor = 0,
     4: binary               charset_name,
     5: binary               collate_name,
-    6: ColumnTypeDef        vid_type,
+    6: ColumnTypeDef        vid_type = {"type": PropertyType.FIXED_STRING, "type_length": 8},
 }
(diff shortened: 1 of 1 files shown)


src/common/interface/meta.thrift
```diff
@@ -173,7 +173,7 @@ struct SpaceDesc {
     3: i32                  replica_factor = 0,
     4: binary               charset_name,
     5: binary               collate_name,
-    6: ColumnTypeDef        vid_type,
+    6: ColumnTypeDef        vid_type = {"type": PropertyType.FIXED_STRING, "type_length": 8},
```
darionyaphet, line 176: Is it a little small?

laura-ding (approved):

src/common/interface/meta.thrift
yixinglu: It's compatible with integer vid type

jievince (approved):

src/common/interface/meta.thrift
darionyaphet: what you defined is the size of string No matter with integer

PR #248, opened 2020-09-23
move reporterror to exprvisitor

nevermore3:


yixinglu (commented):
It's not proper to add this status field and reportError function in the expr visitor interface. Not each error is semantic error. If you need a status, please add your specific visitor implementation.

And do we need a name for each visitor ?

PR #249, opened 2020-09-24
Add the parser to parse time.

Shylock-Hg:


src/common/time/TimeParser.cpp
```diff
@@ -0,0 +1,302 @@
...
+             return Status::Error("Unexpected token `%s'.", toString(n.type));
+         }
+     }},
+    {kDateDay,
+     [](Token t, Token n, DateTime& val, ExpectType type) -> StatusOr<State> {
+         if (t.type == TokenType::kDateDelimiter) {
+             return kDateDay;
+         }
+         if (t.type == TokenType::kNumber) {
+             if (t.val < 1 || t.val > 31) {
```
laura-ding, line 87: Inaccurate judgment, need to judge with the month.

Shylock-Hg: It's only limit checking, when create date will check the date validation.

darionyaphet (approved):

monadbobo:
There are two problems:
1. does our current timestamp not support integers anymore?
2. if it's a string, which ISO standard are we parse according to?

Shylock-Hg:
> There are two problems:
> 
> 1. does our current timestamp not support integers anymore?
> 2. if it's a string, which ISO standard are we parse according to?

1. The pr don't affect the timestamp
2. Follow the opencypher

src/common/time/TimeParser.h
```diff
@@ -0,0 +1,133 @@
...
+        TokenType type;
+        int32_t val;   // only used for number token
+        std::string str;
+    };
+
+    static std::string toString(const Token& t);
+
+    // TODO(shylock) support weeks/days format when support the convertion
+    // The state of current parser
+    enum State {
```
monadbobo, line 85: enum class.

src/common/time/TimeParser.cpp
```diff
@@ -0,0 +1,439 @@
...
+                try {
+                    t.val = folly::to<int32_t>(digits);
+                } catch (std::exception &e) {
+                    return Status::Error("%s", e.what());
+                }
+                t.type = TokenType::kNumber;
+                tokens_.emplace_back(t);
+                digits.clear();
+            }
+        } else if (kTimeDelimiter == *c) {
```
monadbobo, line 391: Maybe it's more intuitive to use switch cases?

src/common/time/TimeParser.cpp
```diff
@@ -0,0 +1,439 @@
...
+#include "common/time/TimezoneInfo.h"
+
+namespace nebula {
+namespace time {
+
+/*static*/ const std::vector<TimeParser::Component> TimeParser::dateTimeStates = {
+    // State, ShiftMap
+    {kInitial,
+     [](Token, Token, Context &ctx) -> StatusOr<State> {
+         if (ctx.type == ExpectType::kDateTime) {
```
monadbobo, line 19: Would it be better to use a hashmap here?

src/common/time/TimeParser.h
Shylock-Hg: This used for array index.

src/common/time/TimeParser.cpp
Shylock-Hg: There is a condition `isdigit`.

src/common/time/test/TimeParserTest.cpp
```diff
@@ -0,0 +1,249 @@
...
+
+#include "common/base/Base.h"
+#include "common/time/TimeParser.h"
+
+TEST(TimeParser, DateTime) {
+    {
+        nebula::time::TimeParser parser;
+        auto result = parser.parseDateTime("2019-01-03T22:22:3.2333");
+        ASSERT_TRUE(result.ok()) << result.status();
+        EXPECT_EQ(nebula::DateTime(2019, 1, 3, 22, 22, 3, 2333), result.value());
```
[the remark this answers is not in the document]
Shylock-Hg: Yes it's just the utc time.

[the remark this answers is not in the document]
Shylock-Hg: Got it. The process of fraction of second had be corrected.

src/common/time/TimeParser.cpp
```diff
@@ -0,0 +1,418 @@
...
+
+#include "common/time/TimeParser.h"
+#include <sstream>
+#include "common/time/TimeUtils.h"
+#include "common/time/TimezoneInfo.h"
+
+namespace nebula {
+namespace time {
+
+/*static*/ const std::vector<TimeParser::Component> TimeParser::dateTimeStates = {
```
yixinglu, line 15: please try to use FSM to solve these states transfer issue, that's powerful tool for string parser, need not any branches.

Shylock-Hg: I'm not sure about your meaning. Could you detail the structure about the FSM you means, and how to replace the branches.

src/common/datatypes/Date.cpp
```diff
@@ -101,21 +106,25 @@ std::string Date::toString() const {
...
 std::string Time::toString() const {
+    auto microsecStr = folly::stringPrintf("%.9f", static_cast<uint32_t>(microsec) / 1000000.0);
+    auto decimalPart = decimal(microsecStr);
     // It's in current timezone already
-    return folly::stringPrintf("%02d:%02d:%02d.%06d", hour, minute, sec, microsec);
+    return folly::stringPrintf("%02d:%02d:%02d%s", hour, minute, sec, decimalPart.c_str());
 }
 
 std::string DateTime::toString() const {
+    auto microsecStr = folly::stringPrintf("%.9f", static_cast<uint32_t>(microsec) / 1000000.0);
```
Aiee, line 116: Why do we keep 9 digits for ms instead of 6? Nebula's time precision is different from openCypher.

Shylock-Hg: What's the difference? They are both fraction of second.

Shylock-Hg:
Move to https://github.com/vesoft-inc/nebula/pull/2770

PR #259, opened 2020-10-14
Get tag vertices or edgetype's edges

panda-sheep:
This pr is based on pr #228
Need to merge pr #228 first

### DIFF
--- a/src/common/interface/meta.thrift
+++ b/src/common/interface/meta.thrift
@@ -206,11 +206,12 @@ union SchemaID {
 }
 
 struct IndexItem {
-    1: common.IndexID      index_id,
-    2: binary              index_name,
-    3: SchemaID            schema_id
-    4: binary              schema_name,
-    5: list<ColumnDef>     fields,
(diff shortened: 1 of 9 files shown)


src/common/interface/meta.thrift
```diff
@@ -574,12 +575,22 @@ struct HBReq {
     5: binary     git_info_sha
 }
 
+enum IndexType {
+    ALL            = 0x01, // all tag indextype or all edge indextype
+    NORMAL         = 0x02,
+    VERTEX         = 0x03,
```
darionyaphet, line 581: what different between NORMAL and VERTEX?

panda-sheep: emm， it is my negligence, I will add notes to it later.
NORMAL  is  previous index on tag or edge, suitable for lookup.
VERTEX  is the index on the tag, which is used to stattistics the count of vertice on the tag 
or  stattistics all vertices on the tag

panda-sheep:
Reimplement, close this pr.

PR #260, opened 2020-10-15
Support text search.

bright-starry-sky:


bright-starry-sky:
added create|drop index for meta.

src/common/plugin/fulltext/FTMetaAdapter.h
```diff
@@ -0,0 +1,42 @@
...
+    virtual ~FTMetaAdapter() = default;
+
+    virtual bool createIndex(const HttpClient& client,
+                             const std::string& index,
+                             const std::string& indexTemplate = "") const = 0;
+
+    virtual bool dropIndex(const HttpClient& client,
+                           const std::string& index) const = 0;
+
+    virtual bool activeNode(const HttpClient& client,
```
critical27, line 29: Is it used for checking liveness?

src/common/plugin/fulltext/FTUtils.h
```diff
@@ -0,0 +1,179 @@
...
+namespace plugin {
+
+enum class FT_SEARCH_OP {
+    kPrefix = 0,
+    kWildcard = 1,
+    kRegexp = 2,
+    kFuzzy = 3,
+};
+
+struct HttpClient {
```
critical27, line 34: Could we reuse the `HttpClient` in common?

src/common/plugin/fulltext/FTMetaAdapter.h
bright-starry-sky: > Is it used for checking liveness?

Yeah, this is the reserved interface for checking valid nodes in the full text cluster

src/common/plugin/fulltext/FTUtils.h
bright-starry-sky: > Could we reuse the `HttpClient` in common?

Emm, This module will be compiled into a separate dynamic library for plugin. I don't want it to related on too many modules, so it's isolated here. Would you like to compile library thrift into this library as well?

src/common/clients/meta/MetaClient.cpp
```diff
@@ -172,6 +172,11 @@ bool MetaClient::loadData() {
         return false;
     }
 
+    if (!loadFulltextClients()) {
```
darionyaphet, line 175: Seems Full Text Search is not an optional ?

src/common/plugin/fulltext/elasticsearch/ESGraphAdapter.cpp
```diff
@@ -0,0 +1,296 @@
...
+
+bool ESGraphAdapter::createIndex(const HttpClient& client,
+                                 const std::string& index,
+                                 const std::string&) const {
+    // curl -H "Content-Type: application/json; charset=utf-8"
+    //      -XPUT "http://127.0.0.1:9200/index_exist"
+    std::string cmd = createIndexCmd(client, index);
+    auto ret = nebula::ProcessUtils::runCommand(cmd.c_str());
+    if (!ret.ok()) {
+        LOG(ERROR) << "Http PUT Failed: " << cmd;
```
darionyaphet, line 212: Here is `Create Index Failed` ? Not the HTTP transform failed

src/common/interface/meta.thrift
```diff
@@ -921,6 +922,33 @@ struct ListGroupsResp {
...
+    ELASTICSEARCH = 0x01,
+} (cpp.enum_strict)
+
+struct FTClient {
+    1: required common.HostAddr    host,
+    2: optional binary             user,
+    3: optional binary             pwd,
+}
+
+struct SignInFTServiceReq {
```
critical27, line 935: Let's discuss the interface of `listener` here. How about `ADD LISTENER ip:port of space space_name`? WDYT?

src/common/clients/meta/MetaClient.cpp
bright-starry-sky: > Seems Full Text Search is not an optional ?

yes, the optional in spaceDesc

src/common/interface/meta.thrift
bright-starry-sky: > Let's discuss the interface of `listener` here. How about `ADD LISTENER [LISTENER_TYPE] list<ip:port> of space`? WDYT?
**I was added an option(TEXT_SEARCH=bool) in create space sentence. it saved into `spaceDesc`. What do you think?
for example CREATE SPACE space1(partition_num=9, replica_factor=1, text_search=true );**
> 
> Something like this:
> 
> 1. Add listener
> 2. After `add listener` is called, meta will allocate the listener of different part on those hosts. (need to save these information on meta)
> 3. The process on those hosts will find that it has added new listeners of some parts, then begin to initialize and join the raft group.
> 
> There are two ways to save these information:
> 
> 1. key: listener host, value: space + list<part + type>
> 2. key: space + part + type, value: host
> 
> Which one do you prefer?
**about 2, Is currently active listener host of `value: host`? if yes. I like 2.**

critical27: 1. The problem of `CREATE SPACE space1(partition_num=9, replica_factor=1, text_search=true );` is that, it will be hard to tell where is the listener of each partition, because it will be distributed among all active hosts. And it will be harder when we add a new type of listener, the `spaceDesc` need to add more and more field.

2. The both ways to save listener information look ok to me. The first way is easier to load and init, because as for a host, it can simply grab all listener of it. The second way will be more friendly.

src/common/plugin/fulltext/elasticsearch/ESGraphAdapter.cpp
```diff
@@ -0,0 +1,278 @@
...
+                    }
+                } else {
+                    continue;
+                }
+            }
+        }
+        return true;
+    }
+    LOG(ERROR) << "error reason : " << ret;
+    LOG(ERROR) << ret;
```
darionyaphet, line 39: why logging ret double?

bright-starry-sky: > why logging ret double?

Good point，my fault.

src/common/plugin/fulltext/FTStorageAdapter.h
```diff
@@ -0,0 +1,32 @@
...
+namespace nebula {
+namespace plugin {
+
+class FTStorageAdapter {
+public:
+    virtual ~FTStorageAdapter() = default;
+
+    virtual StatusOr<bool> put(const HttpClient& client, const DocItem& item) const = 0;
+
+    virtual StatusOr<bool> bulk(const HttpClient& client,
```
critical27, line 22: After digging a little in storage,  there comes one question: since bulk need to support both put and delete in one batch, so I guess the DocItem need one more parameter to know it is put or delete?

The process would be `parse wal` -> `decode into kv with operation type` -> DocItem.

bright-starry-sky: > After digging a little in storage, there comes one question: since bulk need to support both put and delete in one batch, so I guess the DocItem need one more parameter to know it is put or delete?
> 
> The process would be `parse wal` -> `decode into kv with operation type` -> `DocItem` -> `http request`?

We don't need to handler about delete data for es. allow es to have redundant data.

bright-starry-sky:
Ready to review.

src/common/interface/meta.thrift
critical27: Do we still need these methods?

bright-starry-sky: I think so. SignInFTServiceReq used to handle Elasticsearch clients. Listener used to handle `kv` nodes. I think it's better to be independent. WDYT?

src/common/plugin/fulltext/elasticsearch/ESGraphAdapter.cpp
```diff
@@ -0,0 +1,277 @@
...
+        return Status::Error("command failed : %s", cmd.c_str());
+    }
+    return indexCheck(ret.value());
+}
+
+std::string ESGraphAdapter::indexExistsCmd(const HttpClient& client,
+                                           const std::string& index) const noexcept {
+    std::stringstream os;
+    os << header() << XGET << client.toString()
+    << "_cat/indices/" << index << "?format=json" << "\"";
```
darionyaphet, line 254: alignment ?

bright-starry-sky: > alignment ?

Good point. thanks.

bright-starry-sky:
integrate with listener.

src/common/plugin/fulltext/elasticsearch/ESGraphAdapter.cpp
```diff
@@ -0,0 +1,277 @@
...
+    folly::dynamic value = folly::dynamic::object("value", items);
+    return folly::dynamic::object("match", value);
+}
+
+
+StatusOr<bool> ESGraphAdapter::createIndex(const HttpClient& client,
+                                           const std::string& index,
+                                           const std::string&) const {
+    // curl -H "Content-Type: application/json; charset=utf-8"
+    //      -XPUT "http://127.0.0.1:9200/index_exist"
```
critical27, line 200: Better rename all `index_exist`, it seems is nebula_spaceName_tag/edge according to https://github.com/vesoft-inc/nebula-graph/pull/365

bright-starry-sky:
Done. ready to review.

src/common/plugin/fulltext/FTGraphAdapter.h
```diff
@@ -0,0 +1,57 @@
...
+namespace nebula {
+namespace plugin {
+
+class FTGraphAdapter {
+public:
+    FTGraphAdapter() = default;
+
+    virtual ~FTGraphAdapter() = default;
+
+    virtual StatusOr<bool> prefix(const HttpClient& client,
```
critical27, line 22: return Status is enough?

src/common/interface/meta.thrift
```diff
@@ -981,6 +984,33 @@ struct GetStatisResp {
...
```
critical27, line 997: After a rethink, what do you think of use a `gflag` instead of adding more sing in/out?

src/common/plugin/fulltext/elasticsearch/ESGraphAdapter.h
```diff
@@ -0,0 +1,108 @@
...
+    FRIEND_TEST(FulltextPluginTest, ESWildcardTest);
+    FRIEND_TEST(FulltextPluginTest, ESRegexpTest);
+    FRIEND_TEST(FulltextPluginTest, ESFuzzyTest);
+    FRIEND_TEST(FulltextPluginTest, ESCreateIndexTest);
+    FRIEND_TEST(FulltextPluginTest, ESDropIndexTest);
+
+public:
+    static std::unique_ptr<FTGraphAdapter> kAdapter;
+
+    StatusOr<bool> prefix(const HttpClient& client,
```
critical27, line 28: It seems all these interfaces return `Status` is enough?

src/common/plugin/fulltext/FTGraphAdapter.h
bright-starry-sky: > return Status is enough?

Emm.. I don't think so. Three types of errors may be returned：

1. Status::ERROR : means http error or FT api error. retry needed if this error occurs.
2. true : Execute successfully
3. false : Elasticsearch internal error, which can be a result set parse error. not need retry.

src/common/interface/meta.thrift
bright-starry-sky: > After a rethink, what do you think of use a `gflag` instead of adding more sing in/out?

Good question. Both way are OK.I just think meta-management is more user friendly. If using gflag, need to add this parameter to both storage and graph, and they must be made available.

src/common/plugin/fulltext/FTGraphAdapter.h
critical27: Got it

bright-starry-sky:
Resolve the conflict

src/common/plugin/fulltext/FTUtils.h
```diff
@@ -0,0 +1,209 @@
...
+    HostAddr host;
+    std::string user;
+    std::string password;
+
+    HttpClient() = default;
+    ~HttpClient() = default;
+
+    HttpClient(HttpClient&& v) noexcept
```
darionyaphet, line 43: explicit ?

src/common/plugin/fulltext/FTUtils.h
```diff
@@ -0,0 +1,209 @@
...
+struct DocItem {
+    std::string index;
+    std::string column;
+    int32_t part;
+    int32_t schema;
+    std::string val;
+
+    ~DocItem() = default;
+
+    DocItem(DocItem&& v) noexcept
```
darionyaphet, line 101: ditto

src/common/plugin/fulltext/FTUtils.h
bright-starry-sky: Good point. fixed.

bright-starry-sky:
Addressed darionyaphet's comments.

liuyu85cn (changes requested):

src/common/clients/meta/MetaClient.cpp
```diff
@@ -449,6 +454,22 @@ bool MetaClient::loadListeners(GraphSpaceID spaceId, std::shared_ptr<SpaceInfoCa
...
+     auto ftRet = listFTClients().get();
+     if (!ftRet.ok()) {
+         LOG(ERROR) << "List fulltext services failed, status:" << ftRet.status();
+         return false;
+     }
+     decltype(fulltextClientList_)    serviceClients;
+     serviceClients = ftRet.value();
+     {
+         folly::RWSpinLock::WriteHolder holder(localCacheLock_);
+         fulltextClientList_ = std::move(serviceClients);
```
liuyu85cn, line 467: fulltextClientList_.swap(serviceClients) ?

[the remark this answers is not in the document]
bright-starry-sky: > fulltextClientList_.swap(serviceClients) ?

What is the advantage? I don't think need to `swap`, `move` is enough.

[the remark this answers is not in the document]
bright-starry-sky: > > fulltextClientList_.swap(serviceClients) ?
> 
> What is the advantage? I don't think need to `swap`, `move` is enough.

Your comments led me to other improvements

bright-starry-sky:
Simplified the method `loadFulltextClients`

liuyu85cn (dismissed):

src/common/meta/Common.h
```diff
@@ -63,6 +63,16 @@ using RemoteListeners =
...
+// ListenersMap is used for listener replica to get its peers of data replica
+using ListenersMap =
+    std::unordered_map<GraphSpaceID, std::unordered_map<PartitionID, std::vector<ListenerHosts>>>;
+// RemoteListenerInfo is pair of <listener host, listener type>
+using RemoteListenerInfo = std::pair<HostAddr, cpp2::ListenerType>;
+// RemoteListeners is used for data replica to check if some part has remote listener
+using RemoteListeners =
+    std::unordered_map<GraphSpaceID,
+                       std::unordered_map<PartitionID, std::vector<RemoteListenerInfo>>>;
+
```
critical27, line 75: This could be removed. Previous PR has merged.

critical27 (approved):
LGTM

src/common/plugin/fulltext/elasticsearch/ESGraphAdapter.cpp
```diff
@@ -0,0 +1,299 @@
...
+    const std::string& index, const std::string&) const noexcept {
+    std::stringstream os;
+    os << header() << XPUT << client.toString() << index << "\"";
+    return os.str();
+}
+
+StatusOr<bool> ESGraphAdapter::dropIndex(const HttpClient& client,
+                                         const std::string& index) const {
+    // curl -H "Content-Type: application/json; charset=utf-8"
+    //      -XDELETE "http://127.0.0.1:9200/index_exist"
```
panda-sheep, line 194: Use this interface to delete data in es?

src/common/plugin/fulltext/elasticsearch/ESGraphAdapter.h
```diff
@@ -0,0 +1,108 @@
+/* Copyright (c) 2020 vesoft inc. All rights reserved.
+ *
+ * This source code is licensed under Apache 2.0 License,
+ * attached with Common Clause Condition 1.0, found in the LICENSES directory.
+ */
+
+#ifndef NEBULA_PLUGIN_ESGRAPHADAPTER_H
```
panda-sheep, line 7: `COMMON_PLUGIN_ESGRAPHADAPTER_H `   is better？

src/common/plugin/fulltext/FTUtils.h
```diff
@@ -0,0 +1,209 @@
...
+
+enum class FT_SEARCH_OP {
+    kPrefix = 0,
+    kWildcard = 1,
+    kRegexp = 2,
+    kFuzzy = 3,
+};
+
+struct HttpClient {
+    HostAddr host;
```
panda-sheep, line 36: The naming of member variables ends with an underscore
Same as below

src/common/plugin/fulltext/FTUtils.h
```diff
@@ -0,0 +1,209 @@
...
+        return os.str();
+    }
+};
+
+struct DocItem {
+    std::string index;
+    std::string column;
+    int32_t part;
+    int32_t schema;
+    std::string val;
```
panda-sheep, line 97: ditto

src/common/plugin/fulltext/elasticsearch/ESStorageAdapter.cpp
```diff
@@ -0,0 +1,217 @@
...
+        auto root = folly::parseJson(ret);
+        auto result = root.find("errors");
+        if (result != root.items().end() && result->second.isBool() && !result->second.getBool()) {
+            return true;
+        }
+    } catch (std::exception &e) {
+        LOG(ERROR) << "result error : " << e.what();
+    }
+    VLOG(3) << "Bulk insert failed";
+    VLOG(3) << ret;
```
panda-sheep, line 117: `VLOG(3) << "Bulk insert failed" << ret;  `   is better?

src/common/plugin/fulltext/test/FulltextPluginTest.cpp
```diff
@@ -0,0 +1,351 @@
...
+    HostAddr localHost_{"127.0.0.1", 9200};
+    HttpClient client(localHost_);
+    auto ret = ESGraphAdapter().indexExistsCmd(client, "test_index");
+    auto expected = "/usr/bin/curl -H \"Content-Type: application/json; charset=utf-8\" "
+                    "-XGET \"http://127.0.0.1:9200/_cat/indices/test_index?format=json\"";
+    ASSERT_EQ(expected, ret);
+}
+
+TEST(FulltextPluginTest, ESCreateIndexTest) {
+    HostAddr localHost_{"127.0.0.1", 9200};
```
panda-sheep, line 28: If there is no es server on 127.0.0.1, can this UT pass the test?

panda-sheep (commented):
Good job， Excellent👍

src/common/plugin/fulltext/elasticsearch/ESGraphAdapter.cpp
bright-starry-sky: > Use this interface to delete data in es?

No, used drop es index.

src/common/plugin/fulltext/elasticsearch/ESGraphAdapter.h
bright-starry-sky: > `COMMON_PLUGIN_ESGRAPHADAPTER_H ` is better？

the name space is nebula::plugin, I think the `NEBULA_PLUGIN_ESGRAPHADAPTER_H` is ok.

src/common/plugin/fulltext/elasticsearch/ESStorageAdapter.cpp
bright-starry-sky: No, here need a '\n'. because a separate display is required here.

src/common/plugin/fulltext/test/FulltextPluginTest.cpp
bright-starry-sky: > If there is no es server on 127.0.0.1, can this UT pass the test?

There's no real call to ES here. please review method `createIndexCmd`

src/common/plugin/fulltext/FTUtils.h
bright-starry-sky: > The naming of member variables ends with an underscore
> Same as below

It's just a personal habit. I see that struct `Edge` doesn't end with an underscore.

src/common/clients/meta/MetaClient.cpp
```diff
@@ -3224,9 +3255,70 @@ MetaClient::getStatis(GraphSpaceID spaceId) {
...
+                },
                 std::move(promise));
     return future;
 }
 
+
+StatusOr<std::vector<cpp2::FTClient>> MetaClient::getFTClientsFromCache() {
+    if (!ready_) {
+        return Status::Error("Not ready!");
+    }
```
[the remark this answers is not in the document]
bright-starry-sky: > in line 464，use write lock of `localCacheLock_`， here should use read lock of `localCacheLock_`?

I don't think so. There is only one line of code to return the result set. What's the point of using a lock?

src/common/meta/Common.h
bright-starry-sky: > This could be removed. Previous PR has merged.

Good point. What did I say , there's a real merger problem. thanks. removed.

src/common/expression/TextSearchExpression.cpp
```diff
@@ -0,0 +1,78 @@
...
+        case Kind::kTSPrefix: {
+            buf = "PREFIX(";
+            break;
+        }
+        case Kind::kTSFuzzy: {
+            buf = "FUZZY(";
+            break;
+        }
+        case Kind::kTSRegexp: {
+            buf = "FUZZY(";
```
darionyaphet, line 67: `Regexp` ?

bright-starry-sky: Good point. fixed. thanks.

src/common/plugin/fulltext/elasticsearch/ESGraphAdapter.h
panda-sheep: Because the common code of the common repo is moved to the common directory, it is changed to COMMON_XX, please refer to other files

src/common/clients/meta/MetaClient.cpp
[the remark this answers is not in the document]
panda-sheep: My mistake, you are right

src/common/plugin/fulltext/test/FulltextPluginTest.cpp
panda-sheep: Got， thx

liuyu85cn (approved):

critical27 (approved):
Good job

PR #261, opened 2020-10-16
Implement fulltext service for meta client

bright-starry-sky:
in progress. waiting storage side.

### DIFF
--- a/src/common/interface/meta.thrift
+++ b/src/common/interface/meta.thrift
@@ -920,6 +920,62 @@ struct ListGroupsResp {
     3: list<Group>      groups,
 }
 
+struct FTHost {
+    1: required common.HostAddr    host,
+    2: optional binary             user,
+    3: optional binary             pwd,
+}
+
+enum FTIndexType {
+    TAG    = 0x01,
+    EDGE   = 0x02,
+} (cpp.enum_strict)
+
+struct FTIndexItem {
+    1: required binary              index_name,
+    2: optional binary              index_template,
+    3: required FTIndexType         index_type,
+}
+
+struct AddFTHostsReq {
+    1: binary            name,
+    2: list<FTHost>      hosts,
+}
+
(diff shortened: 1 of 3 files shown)


bright-starry-sky:
Ready to review. thanks.

src/common/interface/meta.thrift
```diff
@@ -826,6 +826,61 @@ struct ListIndexStatusResp {
...
+} (cpp.enum_strict)
+
+struct FTIndexItem {
+    1: required binary              index_name,
+    2: optional binary              index_template,
+    3: required FTIndexType         index_type,
+}
+
+struct AddFTHostsReq {
+    1: binary            name,
```
critical27, line 847: What is the usage of `name` field in. `AddFTHostsReq` and `RemoveFTHostsReq`? Is it an alias of hosts?

bright-starry-sky: > What is the usage of `name` field in. `AddFTHostsReq` and `RemoveFTHostsReq`? Is it an alias of hosts?

Used to register or unregister the full-text cluster, the name means cluster name. hosts means list of fulltext instance (ip:port) .

critical27: Ah, got it.

critical27: If so, do we need to specify `name` when create/drop full text index?

bright-starry-sky: > If so, do we need to specify `name` when create/drop full text index?

Good question, we need to specify `name` when create full text index. There may be two indexes in a space，tag index and edge index. We need to use this 'name' to identify them.

I also think the name 'AddFTHostsReq' is ambiguous,  is better using `signInFTService` and `signOutFTService` ?

bright-starry-sky:
Addressed  @critical27's comments.

bright-starry-sky:
Close this pr. unified code to #260

PR #263, opened 2020-10-19
add getSpaceVidType

critical27:
To fix https://github.com/vesoft-inc/nebula-storage/issues/170 and https://github.com/vesoft-inc/nebula-storage/issues/166, storage need to be aware of vid type. And can trim trailing \0 if the space vid is string with fixed length.

Needed by https://github.com/vesoft-inc/nebula-storage/pull/172.


bright-starry-sky (approved):
LGTM. thanks for your take this.

panda-sheep (approved):
Good job!

PR #267, opened 2020-10-21
Add builtin functions on Edge/Vertex

dutor:
As title.


src/common/expression/AttributeExpression.cpp
```diff
@@ -29,6 +33,18 @@ const Value& AttributeExpression::eval(ExpressionContext &ctx) {
         }
         return Value::kNullValue;
     } else if (lvalue.isEdge()) {
+        if (rvalue.getStr()[0] == '_') {
```
yixinglu, line 36: Is the `rvalue.getStr()` always not empty ?  Is it safer to check the condition by `rvalue.getStr().find('_') == 0`?

src/common/expression/AttributeExpression.cpp
```diff
@@ -21,6 +21,10 @@ const Value& AttributeExpression::eval(ExpressionContext &ctx) {
     if (lvalue.isMap()) {
         return lvalue.getMap().at(rvalue.getStr());
     } else if (lvalue.isVertex()) {
+        if (rvalue.getStr() == "_vid") {
```
laura-ding, line 24: use `kVid`

src/common/expression/AttributeExpression.cpp
laura-ding: The Parser guarantees that the incoming cannot be empty. if use `find`, it can return the wrong value because the prop name can include '_' but can’t begin.

But it's better to check the length.

src/common/function/FunctionManager.cpp
```diff
@@ -771,6 +787,91 @@ FunctionManager::FunctionManager() {
...
+        attr.maxArity_ = 1;
+        attr.body_ = [](const auto &args) -> Value {
+            if (args[0].type() != Value::Type::VERTEX) {
+                return Value::kNullBadType;
+            }
+            List tags;
+            for (auto &tag : args[0].getVertex().tags) {
+                tags.emplace_back(tag.name);
+            }
+            return tags;
```
laura-ding, line 828: define the same body with "tags"

src/common/expression/AttributeExpression.cpp
```diff
@@ -29,6 +33,18 @@ const Value& AttributeExpression::eval(ExpressionContext &ctx) {
...
+        if (rvalue.getStr()[0] == '_') {
+            if (rvalue.getStr() == "_src") {
+                result_ = lvalue.getEdge().src;
+            } else if (rvalue.getStr() == "_dst") {
+                result_ = lvalue.getEdge().dst;
+            } else if (rvalue.getStr() == "_rank") {
+                result_ = lvalue.getEdge().ranking;
+            } else if (rvalue.getStr() == "_type") {
+                result_ = lvalue.getEdge().type;
+            }
```
CPWstatic, line 45: kSrc,kDst,kRank,kType
Besides, we need some tests.

src/common/expression/AttributeExpression.cpp
Shylock-Hg: I don't think this only used in parser, so better add a debug check.

dutor: @yixinglu 
It's definitely a bug if a label is empty. I will add a assertion to make it clear.

src/common/expression/AttributeExpression.cpp
dutor: thanks!

src/common/function/FunctionManager.cpp
dutor: agree.

src/common/expression/AttributeExpression.cpp
dutor: added. thanks

CPWstatic (approved):

src/common/function/FunctionManager.cpp
```diff
@@ -771,6 +790,97 @@ FunctionManager::FunctionManager() {
             }
         };
     }
+    {
+        auto &attr = functions_["id"];
+        attr.minArity_ = 1;
+        attr.maxArity_ = 1;
+        attr.body_ = [](const auto &args) -> Value {
+            if (args[0].type() != Value::Type::VERTEX) {
```
laura-ding, line 798: !args[0].isVertex(), under ditto

src/common/function/FunctionManager.cpp
```diff
@@ -131,6 +133,23 @@ std::unordered_map<std::string, std::vector<TypeSignature>> FunctionManager::typ
               TypeSignature({Value::Type::SET}, Value::Type::INT),
               TypeSignature({Value::Type::DATASET}, Value::Type::INT),
              }},
+    {"id", {TypeSignature({Value::Type::VERTEX}, Value::Type::STRING),
```
laura-ding, line 136: here can be int

src/common/function/FunctionManager.cpp
```diff
@@ -131,6 +133,23 @@ std::unordered_map<std::string, std::vector<TypeSignature>> FunctionManager::typ
...
+    {"tags", {TypeSignature({Value::Type::VERTEX}, Value::Type::LIST),
+             }},
+    {"labels", {TypeSignature({Value::Type::VERTEX}, Value::Type::LIST),
+             }},
+    {"properties", {TypeSignature({Value::Type::VERTEX}, Value::Type::MAP),
+                    TypeSignature({Value::Type::EDGE}, Value::Type::MAP),
+             }},
+    {"type", {TypeSignature({Value::Type::EDGE}, Value::Type::STRING),
+             }},
+    {"src", {TypeSignature({Value::Type::EDGE}, Value::Type::STRING),
```
laura-ding, line 147: ditto

src/common/expression/test/ExpressionTest.cpp
```diff
@@ -1642,24 +1642,61 @@ TEST_F(ExpressionTest, MapAttribute) {
...
+        auto *left = new ConstantExpression(Value(edge));
         auto *right = new LabelExpression(new std::string("Rocky"));
         AttributeExpression expr(left, right);
         auto value = Expression::eval(&expr, gExpCtxt);
         ASSERT_TRUE(value.isStr());
         ASSERT_EQ("Raccoon", value.getStr());
     }
+    {
+        auto *left = new ConstantExpression(Value(edge));
+        auto *right = new LabelExpression(new std::string("_type"));
```
laura-ding, line 1665: unify to use kType

CPWstatic (approved):

laura-ding (approved):

src/common/expression/ConstantExpression.h
```diff
@@ -27,6 +27,10 @@ class ConstantExpression : public Expression {
         return val_;
     }
 
+    const Value& value() const {
```
Shylock-Hg, line 30: Why add this?

PR #272, opened 2020-10-23
Support rebuild multi index

darionyaphet:


liuyu85cn (dismissed):
LGTM

yixinglu:
How about the relationship between this PR and https://github.com/vesoft-inc/nebula-graph/pull/600 ?

PR #273, opened 2020-10-23
Add unknown job command.

Shylock-Hg:


src/common/interface/meta.thrift
```diff
@@ -286,6 +286,7 @@ enum AdminCmd {
     FLUSH               = 1
     REBUILD_TAG_INDEX   = 2
     REBUILD_EDGE_INDEX  = 3
+    UNKNOWN             = 4
```
darionyaphet, line 289: why support an unknown for admin command?

jievince: I guess it's used for the default value of AdminCmd. Because submitJob() always needs param cmd but some AdminJobOp don't have cmd.

Shylock-Hg: Yes.

darionyaphet (approved):

jievince (approved):

PR #281, opened 2020-10-28
Add the stateless mark to function.

Shylock-Hg:


nevermore3 (approved):

jievince (approved):

dutor (approved):
Thanks!

src/common/function/FunctionManager.h
```diff
@@ -38,6 +38,8 @@ class FunctionManager final {
      */
     static StatusOr<Function> get(const std::string &func, size_t arity);
 
+    static StatusOr<bool> getIsPure(const std::string &func, size_t arity);
```
yixinglu, line 41: it's better to rename this function `isPure`

Shylock-Hg: This is not the method of function but the function manager, so I don't think isPure is better.

jievince (approved):

PR #282, opened 2020-10-28
fixed bug of storage client for index scan

bright-starry-sky:


panda-sheep (approved):
Good，LGTM

src/common/clients/storage/GraphStorageClient.cpp
```diff
@@ -479,7 +479,7 @@ GraphStorageClient::lookupIndex(GraphSpaceID space,
         req.set_return_columns(returnCols);
 
         cpp2::IndexSpec spec;
-        spec.set_contexts(std::move(contexts));
```
Shylock-Hg, line 482: what's the bug?

darionyaphet: const should not move ？

Shylock-Hg: But the origin code is not const.

critical27 (approved):

PR #287, opened 2020-10-29
support regular expression

jievince:
Depends on:
https://github.com/vesoft-inc/nebula-graph/pull/414
https://github.com/vesoft-inc/nebula-storage/pull/217

### DIFF
--- a/src/common/context/ExpressionContext.h
+++ b/src/common/context/ExpressionContext.h
@@ -64,7 +64,19 @@ public:
     // Get Value by Column index
     virtual Value getColumn(int32_t index) const = 0;
 
+    // Get regex
+    const std::regex& getRegex(const std::string& pattern) {
+        auto iter = regex_.find(pattern);
+        if (iter == regex_.end()) {
+            iter = regex_.emplace(pattern, std::regex(pattern)).first;
+        }
+        return iter->second;
+    }
+
     virtual void setVar(const std::string& var, Value val) = 0;
+
+private:
+    std::unordered_map<std::string, std::regex> regex_;
 };
 
 }  // namespace nebula
--- a/src/common/expression/RelationalExpression.cpp
+++ b/src/common/expression/RelationalExpression.cpp
@@ -41,6 +41,20 @@ const Value& RelationalExpression::eval(ExpressionContext& ctx) {
         case Kind::kRelGE:
             result_ = lhs >= rhs;
             break;
+        case Kind::kRelREG: {
+            if (lhs.isStr() && rhs.isStr()) {
+                try {
+                    const auto& r = ctx.getRegex(rhs.getStr());
+                    result_ = std::regex_match(lhs.getStr(), r);
+                } catch (const std::exception& ex) {
+                    LOG(ERROR) << "Regex match error: " << ex.what();
+                    result_ = Value::kNullBadType;
+                }
+            } else {
+                result_ = Value::kNullBadType;
+            }
+            break;
+        }
         case Kind::kRelIn: {
             if (rhs.isList()) {
                 result_ = rhs.getList().contains(lhs);
@@ -147,6 +161,9 @@ std::string RelationalExpression::toString() const {
             break;
         case Kind::kRelNE:
             op = "!=";
+            break;
+        case Kind::kRelREG:
+            op = "=~";
             break;
         case Kind::kRelIn:
             op = " IN ";
--- a/src/common/expression/test/ExpressionContextMock.h
+++ b/src/common/expression/test/ExpressionContextMock.h
@@ -126,5 +126,6 @@ public:
 
 private:
     static std::unordered_map<std::string, Value>      vals_;
+    std::unordered_map<std::string, std::regex>        regex_;
 };
 }  // namespace nebula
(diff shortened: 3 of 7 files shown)


src/common/expression/RelationalExpression.cpp
```diff
@@ -41,6 +42,19 @@ const Value& RelationalExpression::eval(ExpressionContext& ctx) {
         case Kind::kRelGE:
             result_ = lhs >= rhs;
             break;
+        case Kind::kRelREG: {
+            if (lhs.isStr() && rhs.isStr()) {
+                try {
+                    result_ = std::regex_match(lhs.getStr(), std::regex(rhs.getStr()));
```
dutor, line 48: Constructing a `regex` is very expensive. We better to reuse the object in some way.

For your reference:
```
============================================================================
RegexBenchmark.cpp                              relative  time/iter  iters/s
============================================================================
Regex(1_thread)                                            319.74us    3.13K
Regex(2_thread)                                  100.05%   319.59us    3.13K
Regex(4_thread)                                   99.88%   320.12us    3.12K
Regex(8_thread)                                   99.38%   321.73us    3.11K
Regex(16_thread)                                  94.51%   338.33us    2.96K
============================================================================
```

dutor (commented):
Well done.

yixinglu (dismissed):

dutor (commented):
Well done.

src/common/expression/RelationalExpression.cpp
```diff
@@ -41,6 +41,20 @@ const Value& RelationalExpression::eval(ExpressionContext& ctx) {
         case Kind::kRelGE:
             result_ = lhs >= rhs;
             break;
+        case Kind::kRelREG: {
+            if (lhs.isStr() && rhs.isStr()) {
+                try {
+                    auto r = ctx.getRegex(rhs.getStr());
```
dutor, line 47: `auto &r` if `getRegex` returns by reference.

src/common/expression/test/ExpressionContextMock.h
```diff
@@ -117,12 +117,20 @@ class ExpressionContextMock final : public ExpressionContext {
...
+    }
+
     void setVar(const std::string& var, Value val) override {
         UNUSED(var);
         UNUSED(val);
     }
 
 private:
     static std::unordered_map<std::string, Value>      vals_;
+    std::unordered_map<std::string, std::regex>        regex_;
```
Shylock-Hg, line 134: Why need store multiple regex in one regex expression?

jievince: Because one regex may be evaluated multi-time in a query and a query may have multiple regular patterns. So save  compiled regex pattern to Expressioncontext can save much time

Shylock-Hg: I think you'd better store it in expression instead of expression context.

jievince: In a query, a regular pattern may be used mand times. Such as: `where tag.p1 =~ pattern1 and tag.p2=~pattern1`. If stored in expression, pattern1 will be compiled twice

Shylock-Hg: I don't understand, in my mind the `pattern1` and `pattern2` will be stored in different regex expr instance in this case.

yixinglu (approved):
LGTM.

However, in the future we maybe consider to optimize the implementation of this expression about where to save the regex pattern. Now it has following overhead:

1. call virtual function `getRegex` each time in eval
2. find pattern in hash map in ExpressionContext

src/common/context/ExpressionContext.h
```diff
@@ -61,6 +61,9 @@ class ExpressionContext {
     // Get Edge
     virtual Value getEdge() const = 0;
 
+    // Get regex
+    virtual const std::regex& getRegex(const std::string& pattern) = 0;
```
yixinglu, line 65: please add `const` to this interface

src/common/context/ExpressionContext.h
```diff
@@ -62,9 +62,17 @@ class ExpressionContext {
     virtual Value getEdge() const = 0;
 
     // Get regex
-    virtual const std::regex& getRegex(const std::string& pattern) = 0;
+    const std::regex& getRegex(const std::string& pattern) {
+        if (regex_.find(pattern) == regex_.end()) {
+            regex_[pattern] = std::regex(pattern);
```
Shylock-Hg, line 67: Avoid twice lookup.

src/common/context/ExpressionContext.h
```diff
@@ -62,9 +62,17 @@ class ExpressionContext {
...
+        }
+        return regex_[pattern];
```
Shylock-Hg, line 69: ditto.

jievince: Sorry, I was careless

src/common/context/ExpressionContext.h
jievince: It's not const function

nevermore3 (approved):

dutor (approved):
Well done.

PR #303, opened 2020-11-11
Add .debug_gdb_scripts section to improve debugging experience

dutor:
Because we link against libstdc++ statically, GDB cannot automatically load pretty printers for STL data structures. This PR add an option to add a special ELF section to each executable to load those printers.


nevermore3 (approved):
fabulous, Looking forward it

yixinglu (approved):

PR #306, opened 2020-11-16
Enhance path build expression

yixinglu:
Now we can use path builder expression to build the path for following cases:
1. V
2. V, E
3. V, Path
4. Path, V, E
5. Path, Path
6. Path, V, Path

### DIFF
--- a/src/common/expression/test/ExpressionTest.cpp
+++ b/src/common/expression/test/ExpressionTest.cpp
@@ -6,6 +6,7 @@
 
 #include <gtest/gtest.h>
 #include <boost/algorithm/string.hpp>
+#include <memory>
 #include "common/datatypes/DataSet.h"
 #include "common/datatypes/Edge.h"
 #include "common/datatypes/List.h"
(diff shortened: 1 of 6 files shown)


nevermore3 (approved):

src/common/function/FunctionManager.cpp
```diff
@@ -1182,6 +1189,35 @@ FunctionManager::FunctionManager() {
             return result;
         };
     }
+    {
+        auto &attr = functions_["cyclePath"];
```
CPWstatic, line 1193: According to the implementation, the function name should not be named `cyclePath`.  What it really want to do is to find if there exists same edges. So name it like `sameEdgeInPath` or something else properly.

CPWstatic (commented):
LGTM.

src/common/expression/test/ExpressionTest.cpp
```diff
@@ -2732,13 +2781,77 @@ TEST_F(ExpressionTest, PathBuild) {
...
+
+    {
+        auto expr0 = std::make_unique<PathBuildExpression>();
+        expr0->add(varPropExpr("path_src"));
+        auto expr = std::make_unique<PathBuildExpression>();
+        expr->add(std::move(expr0));
+        expr->add(varPropExpr("path_edge1"));
+
+        {
+            // Test: Path + Edge
```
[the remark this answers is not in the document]
yixinglu: in this case,  the expression will check whether the last vertex vid of path and source node of edge is same and the new path will contains a vertex which has no any properties after executing expression eval function.

src/common/function/FunctionManager.cpp
yixinglu: make sense. fixed it.

yixinglu: named `hasSameEdgeInPath`

CPWstatic (approved):

nevermore3 (approved):

PR #307, opened 2020-11-17
Modify binary to string, make client easier to use

laura-ding:
Notice: all client needs to make the same version with the server

Need to modify the ci test for nebula-graph


nevermore3 (approved):

laura-ding:
Do not merge, need to think about the charset how to do.

laura-ding:
close it，because will support multi charset

PR #310, opened 2020-11-18
get data from meta leader

panda-sheep:
For some reasons, the data read from different copies of meta data may be inconsistent.
Therefore, we uniformly obtain data from the leader.


src/common/clients/meta/MetaClient.cpp
```diff
@@ -814,11 +814,14 @@ MetaClient::submitJob(cpp2::AdminJobOp op, cpp2::AdminCmd cmd, std::vector<std::
     req.set_paras(std::move(paras));
     folly::Promise<StatusOr<cpp2::AdminJobResult>> promise;
     auto future = promise.getFuture();
-    getResponse(std::move(req), [] (auto client, auto request) {
+    getResponse(std::move(req),
```
Shylock-Hg, line 817: I think you could change the default argument value to `true` directly.

panda-sheep: make sense

liuyu85cn (approved):
LGTM

darionyaphet:
Raft support strong consistent to make all peer hold the same data why read different peer could found different value ？

Shylock-Hg:
> Raft support strong consistent to make all peer hold the same data why read different peer could found different value ？

Raft only make major consistent, I think so.

critical27:
Good job. Since there are more work than this, for example, we don't need sync anymore, ActiveHostMan and LastUpdateTimeMan perhaps need to change as well. Would you move this PR to 1.0, let me handle 2.0.

darionyaphet:
> > Raft support strong consistent to make all peer hold the same data why read different peer could found different value ？
> 
> Raft only make major consistent, I think so.

yep vote should more than half

critical27 (approved):

Shylock-Hg (approved):
Well done! 
Maybe we could remove the `toLeader` option and I don't think we need the inconsistent meta data that different from data itself.

PR #321, opened 2020-11-25
modify scan interface, unify fetch colname

critical27:
Some improvement in scan interface. Only one tag/edge is supported in one req. Add two flag, `only_latest_version` and `enable_read_from_follower`.

@Nicole00 @laura-ding , If want to set `only_latest_version` to true, `FLAGS_enable_multi_versions` should be true in storage, otherwise, performance will be hurt. By default `FLAGS_enable_multi_versions` is false, so just use the default value for `only_latest_version` is ok.

BTW, the column name in scan is still empty for now. It need to be unified with other interfaces in next PR. 

---
In next PR:

client need to specify the property (both property in key and value) **explicitly**, property in keys will no longer return by default. If no property is specified, all properties in key and value will be returned. This changes will effect all interfaces, including `go/fetch/lookup/scan`. @CPWstatic @yixinglu @bright-starry-sky @Nicole00 

Next PR will not change the thrift interface, but the behavior will be affected. We can do this in December. Query engine, storage and client(scan) need to modify according to it.


src/common/interface/storage.thrift
```diff
@@ -351,7 +351,7 @@ struct GetPropResponse {
     //
     // Each column represents one peoperty. When all properties are returned, the
     //   column name is in the form of
-    //   "<tag_name>:<prop_alias>" or "<edge_type_name>:<prop_alias>"
+    //   "tag_name.prop_alias" or "edge_type_name.prop_alias"
```
yixinglu, line 354: Why not use the same struct definition with LookupResponse ?

critical27: You mean only use one of `GetPropResponse` and `LookupIndexResp`? Not a big difference, bug `GetProp` need to handle `fetch *`.

src/common/interface/storage.thrift
```diff
@@ -614,7 +614,7 @@ struct ScanVertexRequest {
     2: common.PartitionID                   part_id,
     // start key of this block
     3: optional binary                      cursor,
-    4: list<VertexProp>                     return_columns,
+    4: VertexProp                           return_columns,
```
darionyaphet, line 617: 👍

src/common/interface/storage.thrift
```diff
@@ -623,6 +623,12 @@ struct ScanVertexRequest {
     // only return data in time range [start_time, end_time)
     7: optional i64                         start_time,
     8: optional i64                         end_time,
+    9: optional binary                      filter,
+    // when storage enable multi versions and only_latest_version is true, only return latest version.
+    // when storage disable multi versions, just use the default value.
```
darionyaphet, line 628: when multi-version is disabled, only_latest_version is doesn't work？

critical27: It will work, but it will have one more string copy, that is why I said use default value.

darionyaphet: `default value` is for what?

src/common/interface/storage.thrift
```diff
@@ -652,6 +658,12 @@ struct ScanEdgeRequest {
     // only return data in time range [start_time, end_time)
     7: optional i64                         start_time,
     8: optional i64                         end_time,
+    9: optional binary                      filter,
```
laura-ding, line 661: how do you want to use the filter? like the expression? please add the comment

critical27: This is a todo, just the expression in nebula

src/common/interface/storage.thrift
```diff
@@ -652,6 +658,12 @@ struct ScanEdgeRequest {
...
+    // when storage enable multi versions and only_latest_version is true, only return latest version.
+    // when storage disable multi versions, just use the default value.
+    10: bool                                only_latest_version = false,
+    // if set to false, forbid follower read
+    11: bool                                enable_read_from_follower = true,
```
laura-ding, line 666: if the follower couldn't make the data is the lastest, when will use it?

critical27: The default value could be false. IMO, scan don't always need to read from leader. One of the most important reason is that it is not a consistent snapshot read in storage (for example, we need more than one request even in a part to get all data). Also, for olap case, not always the latest data is needed, they will specify a time range to get the data in that range. That's why the default value is true.

Shylock-Hg: I guess some plugin like text search or OLAP engine may try it.

src/common/interface/storage.thrift
critical27: use default value of `only_latest_version`, which is false

src/common/interface/storage.thrift
critical27: What do you mean?

src/common/interface/storage.thrift
darionyaphet: when multi-version is disabled, `only_latest_version` is true or false is no impact?

critical27: Always work right, but it will hurt performance when you set it to true and multi version is disabled.

darionyaphet (approved):

liuyu85cn (approved):
LGTM

src/common/interface/storage.thrift
yixinglu: Yeah. Even if GetProp and Lookup have different implementations, they can also return the same response.

critical27: This looks not so important to me. What really matters is that all the `DataSet` in response must be in same format and query engine don't need to distinguish it from each other. After this merged see the next PR.

PR #323, opened 2020-11-27
unify all storage interface behavior

critical27:
Explicit is better than implicit.

Previously, storage has many default return columns. For example, `fetch tag` will always return vid as first column. `lookup edge` will always return src, rank and dst as first three columns. This will cause many problems, for example, query engine will need to distinguish `DataSet` is from fetch or lookup. Both query engine and storage will have special cases to deal with.

In this PR:

1. *Storage will no more return any default columns. Any columns need to specified explicitly by query engine or client.*

    For example, you can specify `kVid`, `kTag` or any column in tag schema to get the property you want in `VertexProp`. And `kSrc/kType/kRank/kDst` and any column in edge schema in `EdgeProp`. 
    The return column of lookup will be specified in `return_columns` of `LookupIndexRequest`. They also follow above principles.

2. The column name in response is as follows: `tag_name.property` or `edge_name.property`. For example "tag1.col1", "tag1.kVid", "edge1.col3", "edge1.kDst", all column names in `fetch/lookup/scan` is the same. In `go`, the column name a different, the `Dataset` in `go` response is a more complicated one, please see the thrift,  but the property name format is same as other interface.

All interfaces have been unified, including `go`, `fetch`, `lookup`, `update` and `scan`.

### DIFF
--- a/src/common/base/Base.h
+++ b/src/common/base/Base.h
@@ -130,6 +130,7 @@ using VariantType = boost::variant<int64_t, double, bool, std::string>;
 // reserved property names
 constexpr char kId[]    = "_id";
 constexpr char kVid[]   = "_vid";
+constexpr char kTag[]   = "_tag";
 constexpr char kSrc[]   = "_src";
 constexpr char kType[]  = "_type";
 constexpr char kRank[]  = "_rank";
--- a/src/common/interface/storage.thrift
+++ b/src/common/interface/storage.thrift
@@ -465,7 +465,7 @@ struct UpdateVertexRequest {
     4: required common.TagID        tag_id
     5: list<UpdatedProp>            updated_props,
     6: optional bool                insertable = false,
-    // A list of expressions
+    // A list of kSrcProperty expressions, support kVid and kTag
     7: optional list<binary>        return_props,
     // If provided, the update happens only when the condition evaluates true
     8: optional binary              condition,
@@ -484,7 +484,7 @@ struct UpdateEdgeRequest {
     3: EdgeKey                  edge_key,
     4: list<UpdatedProp>        updated_props,
     5: optional bool            insertable = false,
-    // A list of expressions
+    // A list of kEdgeProperty expressions, support kSrc, kType, kRank and kDst
     6: optional list<binary>    return_props,
     // If provided, the update happens only when the condition evaluates true
     7: optional binary          condition,
--- a/src/common/interface/storage.thrift
+++ b/src/common/interface/storage.thrift
@@ -271,7 +271,7 @@ struct GetNeighborsResponse {
     //
     // The value of each tag column is a list of Values, which follows the order of the
     //   property names specified in the above column name. If a vertex does not have
-    //   the specific tag, the value will be NULL
+    //   the specific tag, the value will be empty value
     //
     // Columns after the tag columns are edge columns. One column per edge type. The
     //   column name is in the following format in the same order which specified in EdgeProp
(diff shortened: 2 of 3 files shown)


src/common/base/Base.h
```diff
@@ -130,6 +130,7 @@ using VariantType = boost::variant<int64_t, double, bool, std::string>;
 // reserved property names
 constexpr char kId[]    = "_id";
 constexpr char kVid[]   = "_vid";
+constexpr char kTag[]   = "_tag";
```
bright-starry-sky, line 133: Where will be used with kTag?

critical27: It could be return tagId just like a normal tag property. In some interface, need to return all properties of all tags, for example `fetch *` and `go`, then it could return as one of the property. But all storage interface will support to return it as well.

bright-starry-sky: > It could be return tagId just like a normal tag property. In some interface, need to return all properties of all tags, for example `fetch *` and `go`, then it could return as one of the property. But all storage interface will support to return it as well.

Got, nice.

bright-starry-sky (approved):
LGTM

Shylock-Hg:
> Explicit is better than implicit.
> 
> Previously, storage has many default return columns. For example, `fetch tag` will always return vid as first column. `lookup edge` will always return src, rank and dst as first three columns. This will cause many problems, for example, query engine will need to distinguish `DataSet` is from fetch or lookup. Both query engine and storage will have special cases to deal with.
> 
> In this PR:
> 
> 1. _Storage will no more return any default columns. Any columns need to specified explicitly by query engine or client._
>    For example, you can specify `kVid`, `kTag` or any column in tag schema to get the property you want in `VertexProp`. And `kSrc/kType/kRank/kDst` and any column in edge schema in `EdgeProp`.
>    The return column of lookup will be specified in `return_columns` of `LookupIndexRequest`. They also follow above principles.
> 2. The column name in response is as follows: `tag_name.property` or `edge_name.property`. For example "tag1.col1", "tag1.kVid", "edge1.col3", "edge1.kDst", all column names in `fetch/lookup/scan` is the same. In `go`, the column name a different, the `Dataset` in `go` response is a more complicated one, please see the thrift,  but the property name format is same as other interface.
> 
> All interfaces have been unified, including `go`, `fetch`, `lookup`, `update` and `scan`.

Agreeeeeeeeeeeee.

panda-sheep (approved):
Good job，LGTM！

critical27:
Don't merge plz, wait graph and storage.

src/common/interface/storage.thrift
```diff
@@ -650,31 +624,24 @@ struct ScanEdgeRequest {
...
 struct ScanEdgeResponse {
     1: required ResponseCommon              result,
     // The data will return as a dataset. The format is as follows:
-    // 1. First column is src id
-    // 2. Second column is edge type
-    // 3. Third column is edge rank
-    // 4. Fourth column is dst id
-    // 5. Starting from the fifth column, is the properties specified in request.
-    //    If no_columns is true, there are only the first four columns.
+    // Each column represents one peoperty. the column name is in the form of "tag_name.prop_alias"
```
liuyu85cn, line 643: property

liuyu85cn (dismissed):

src/common/interface/storage.thrift
critical27: good catch.

CPWstatic (approved):

darionyaphet (approved):

Shylock-Hg (approved):
LGTM

liuyu85cn (approved):
LGTM

PR #326, opened 2020-11-30
Update label attribute expression construction and fix empty value computation

yixinglu:
Needed by vesoft-inc/nebula-graph#434

### DIFF
--- a/src/common/expression/LabelAttributeExpression.h
+++ b/src/common/expression/LabelAttributeExpression.h
@@ -7,6 +7,9 @@
 #ifndef COMMON_EXPRESSION_LABELATTRIBUTEEXPRESSION_H_
 #define COMMON_EXPRESSION_LABELATTRIBUTEEXPRESSION_H_
 
+#include <memory>
+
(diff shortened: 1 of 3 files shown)


dutor (approved):

src/common/expression/LabelAttributeExpression.h
```diff
@@ -62,7 +66,7 @@ class LabelAttributeExpression final : public Expression {
 
 private:
     std::unique_ptr<LabelExpression>    lhs_;
-    std::unique_ptr<LabelExpression>    rhs_;
+    std::unique_ptr<ConstantExpression> rhs_;
```
Shylock-Hg, line 69: Why need change this? This is strange.

yixinglu: This is bugfix for label attribute expression access.  When we collect all label expressions, we couldn't make sure which symbol is attribute and which is the real label, for example: in e.likeness label attr expr, only `e` is label, but likeness is not. see my nebula-graph PR.

CPWstatic (approved):

PR #334, opened 2020-12-03
if storage rpc return leader_change, but not set leader host, may core

liuyu85cn:
This may happen if an RPC arrived to a storage, whose leader is not elected.


Shylock-Hg (approved):

Shylock-Hg (approved):

critical27 (approved):
Well done!

PR #337, opened 2020-12-07
Feature/specific lint home

Shylock-Hg:


.linters/cpp/checkKeyword.py
```diff
@@ -0,0 +1,149 @@
...
+    'KW_WHERE',
+    'KW_MATCH',
+    'KW_INSERT',
+    'KW_YIELD',
+    'KW_RETURN',
+    'KW_DESCRIBE',
+    'KW_DESC',
+    'KW_VERTEX',
+    'KW_EDGE',
+    'KW_EDGES',
```
Aiee, line 34: According to the meeting we had last week, should some of these keywords be removed?

PR #340, opened 2020-12-07
getResponse throw not appropriate exception if leader changed

liuyu85cn:
now getResponse() will call itself recursively. 
but in the second level. 
```
auto f = pro.getFuture(); 
```
will throw exception,  because get future from the same promise.


panda-sheep (approved):
Excellent!

Shylock-Hg (approved):
Well done.

PR #344, opened 2020-12-09
Added an API for backup/restore.

monadbobo:


src/common/interface/meta.thrift
```diff
@@ -992,6 +1000,48 @@ struct GetStatisResp {
...
+
+struct HostPair {
+    1: common.HostAddr   from_host,
+    2: common.HostAddr   to_host,
+}
+
+struct RestoreMetaReq {
+    1: list<binary>     files,
+    2: list<HostPair>   hosts,
+}
```
bright-starry-sky, line 1043: Because the db restore is offline, So what is the purpose of hosts here? Is the target file sent via SSH? If so, I'd rather use an external tool. WDYT?

monadbobo: When nebula restore, the meta server is started first.

bright-starry-sky: > When nebula restore, the meta server is started first.

got, Mount the file using `ingest` method . right?

bright-starry-sky:
LGTM, please hold this this PR. waiting for storage ready.

src/common/interface/meta.thrift
monadbobo: > > When nebula restore, the meta server is started first.
> 
> got, Mount the file using `ingest` method . right?

yes.

panda-sheep (approved):
Excellent！

critical27 (approved):

PR #345, opened 2020-12-10
Support predicate expressions

jievince:
https://github.com/vesoft-inc/nebula-graph/pull/467 depends on this pr.
doc: https://neo4j.com/docs/cypher-manual/current/functions/predicate/


Shylock-Hg:
Why not put them into functions?

jievince:
> Why not put them into functions?
To users, they are functions, but in implementation, they cannot be implemented as normal functions, because
1. Arguments of predicate functions are special, such as `all(var in list where filter)`, there are conflicts in parser.yy if define it into function_call_exprssion;
2. neo4j also implements them as expressions.

src/common/expression/PredicateExpression.cpp
```diff
@@ -0,0 +1,173 @@
...
+        result_ = Value::kNullValue;
+        return result_;
+    }
+    if (!listVal.isList()) {
+        result_ = Value::kNullBadType;
+        return result_;
+    }
+    auto& list = listVal.getList();
+
+    if (name == "all") {
```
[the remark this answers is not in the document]
jievince: Sorry,  I can't understand. 
1. Make the variable `name` to be a constexpr? 
2. How can enum variables be strings?

nevermore3 (approved):

CPWstatic (approved):
LGTM.

src/common/expression/PredicateExpression.cpp
```diff
@@ -0,0 +1,194 @@
...
+    std::string buf;
+    buf.reserve(256);
+
+    buf += *name_;
+    buf += "(";
+    buf += *innerVar_;
+    buf += " IN ";
+    buf += collection_->toString();
+    buf += " WHERE ";
+    buf += filter_->toString();
```
yixinglu, line 184: Does this expression always have `WHERE` clause ?

jievince: Yes, it always has `WHERE`

yixinglu (approved):

PR #349, opened 2020-12-14
Support get vertex and edge property in storage client

darionyaphet:
A small change about storage client, support get vertex and edge's property explicitly.

### DIFF
--- a/src/common/clients/storage/GraphStorageClient.cpp
+++ b/src/common/clients/storage/GraphStorageClient.cpp
@@ -165,6 +165,34 @@ GraphStorageClient::addEdges(GraphSpaceID space,
 }
 
 folly::SemiFuture<StorageRpcResponse<cpp2::GetPropResponse>>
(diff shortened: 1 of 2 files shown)


src/common/clients/storage/GraphStorageClient.cpp
```diff
@@ -164,6 +164,34 @@ GraphStorageClient::addEdges(GraphSpaceID space,
...
+GraphStorageClient::getVertexProps(GraphSpaceID space,
+                                   const DataSet& input,
+                                   const std::vector<cpp2::VertexProp>* vertexProps,
+                                   const std::vector<cpp2::Expr>* expressions,
+                                   bool dedup,
+                                   const std::vector<cpp2::OrderBy>& orderBy,
+                                   int64_t limit,
+                                   std::string filter,
+                                   folly::EventBase* evb) {
+    return getProps(space, input, vertexProps, nullptr, expressions,
```
Shylock-Hg, line 177: Seems nothing here.

PR #350, opened 2020-12-15
Rename the ops file to avoid install.

Shylock-Hg:
The client will install all `.h` files so rename to avoid install the ops files.


monadbobo:
Does it affect the use of storage here?

Shylock-Hg:
> Does it affect the use of storage here?

No affect to graph or storage.

monadbobo (approved):
LGTM.

nevermore3 (approved):

dutor (approved):

PR #352, opened 2020-12-16
internal storage client needs to retry using a different port 

liuyu85cn:
as title

### DIFF
--- a/src/common/clients/storage/StorageClientBase.inl
+++ b/src/common/clients/storage/StorageClientBase.inl
@@ -363,7 +363,6 @@ void StorageClientBase<ClientType>::getResponseImpl(
         folly::Promise<StatusOr<Response>> pro,
         std::size_t retry,
         std::size_t retryLimit) {
-    LOG(INFO) << "messi getResponseImpl() retry=" << retry;
     if (evb == nullptr) {
         DCHECK(!!ioThreadPool_);
         evb = ioThreadPool_->getEventBase();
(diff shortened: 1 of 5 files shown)


Shylock-Hg:
This issue caused by the InternalStorageService has the different port from GraphStorageService?

liuyu85cn:
> This issue caused by the InternalStorageService has the different port from GraphStorageService?

yes

Shylock-Hg:
> > This issue caused by the InternalStorageService has the different port from GraphStorageService?
> 
> yes

So why the service don't return itself port.

liuyu85cn:
> > > This issue caused by the InternalStorageService has the different port from GraphStorageService?
> > 
> > 
> > yes
> 
> So why the service don't return it's own port.

if InternalStorageService can't return its own port as leader. 
get response will update leader with that. 

then InternalStorageService and GraphStorageService will race the leader again and again.

critical27 (dismissed):
LGTM

src/common/clients/storage/StorageClientBase.inl
```diff
@@ -246,11 +249,13 @@ StorageClientBase<ClientType>::collectResponse(
                             if (retry < retryLimit && isValidHostPtr(leader)) {
                                 evb->runAfterDelay([this, evb, leader = *leader, r = std::move(r),
                                                     remoteFunc = std::move(remoteFunc), context,
-                                                    start, retry, retryLimit] () {
+                                                    start, retry, retryLimit,
+                                                    portOffsetIfRetry] (){
```
panda-sheep, line 253: Why is the leader's port not changed here?

panda-sheep (approved):
Excellent！

src/common/clients/storage/StorageClientBase.inl
panda-sheep: It has been discussed offline.

critical27 (approved):

PR #361, opened 2020-12-24
Fix drop index

laura-ding:


src/common/clients/meta/MetaClient.cpp
```diff
@@ -1667,7 +1667,7 @@ MetaClient::dropTagIndex(GraphSpaceID spaceID, std::string name, bool ifExists)
                     return client->future_dropTagIndex(request);
                 },
                 [] (cpp2::ExecResp&& resp) -> IndexID {
```
Shylock-Hg, line 1669: See the return type.

laura-ding: fixed

Shylock-Hg (approved):

yixinglu (approved):

PR #363, opened 2020-12-24
Support operator+ of List

jievince:


src/common/datatypes/test/ValueTest.cpp
```diff
@@ -89,6 +94,26 @@ TEST(Value, Arithmetics) {
...
+
+        v = vList1 + vBool2;
+        EXPECT_EQ(Value::Type::LIST, v.type());
+        EXPECT_EQ(List({1, 3, 2, true}), v.getList());
+
+        v = vStr1 + vList2;
+        EXPECT_EQ(Value::Type::LIST, v.type());
+        EXPECT_EQ(List({std::string("Hello "), 6, 4, 5}), v.getList());
+
+        v = vMap + vList1;
```
Shylock-Hg, line 110: A little strange behaviour, maybe add list and scalar type only is better.

jievince: image
cypher supports it

Shylock-Hg: ok

Aiee (dismissed):

src/common/datatypes/Value.cpp
```diff
@@ -1830,6 +1860,100 @@ Value operator+(const Value& lhs, const Value& rhs) {
...
+                case Value::Type::EDGE:
+                case Value::Type::PATH:
+                case Value::Type::MAP:
+                case Value::Type::SET: {
+                    auto ret = lhs.getList();
+                    ret.emplace_back(rhs);
+                    return ret;
+                }
+                default: {
+                    return Value::kNullBadType;
```
Shylock-Hg, line 1898: Could remove the default.

jievince: `DataSet` goes into default branch.

Shylock-Hg: Yes you could add DataSet instead of default.

jievince: When I changed it to default, g++ reports an compile error:\
```
/home/jie.wang/another/nebula-common/src/common/datatypes/Value.cpp: In function ‘nebula::Value nebula::operator+(const nebula::Value&, const nebula::Value&)’:
/home/jie.wang/another/nebula-common/src/common/datatypes/Value.cpp:1874:20: error: enumeration value ‘__EMPTY__’ not handled in switch [-Werror=switch]
 1874 |             switch (rhs.type()) {
      |                    ^
/home/jie.wang/another/nebula-common/src/common/datatypes/Value.cpp:1874:20: error: enumeration value ‘NULLVALUE’ not handled in switch [-Werror=switch]
/home/jie.wang/another/nebula-common/src/common/datatypes/Value.cpp:1900:13: error: this statement may fall through [-Werror=implicit-fallthrough=]
 1900 |             }
```
I think the default branch is needed in `switch case`.

Shylock-Hg: Yes add them too.

jievince: But `__EMPTY__` and `NULLVALUE` have actually been processed at the start of the function `operator+`
```
    if (lhs.isNull() || (lhs.empty() && !rhs.isNull())) {
        return lhs;
    }
    if (rhs.isNull() || rhs.empty()) {
        return rhs;
    }
```

Shylock-Hg: The default is almost same to add the whole enum options but the default will prevent the compiler warning you when new enum options added.

jievince: Thanks, I've fixed it

src/common/datatypes/Value.cpp
```diff
@@ -1830,6 +1860,109 @@ Value operator+(const Value& lhs, const Value& rhs) {
...
+                }
+                case Value::Type::__EMPTY__: {
+                    return Value::kEmpty;
+                }
+                case Value::Type::NULLVALUE: {
+                    return Value::kNullValue;
+                }
+                default: {
+                    LOG(FATAL) << "Unknown type: " << rhs.type();
```
Shylock-Hg, line 1907: You could fatal outside the switch to remove the default branch.

jievince: Good idea!

dutor (approved):

src/common/datatypes/Value.cpp
```diff
@@ -1724,6 +1724,11 @@ Value operator+(const Value& lhs, const Value& rhs) {
                                                lhs.getBool() ? "true" : "false",
                                                rhs.getStr().c_str());
                 }
+                case Value::Type::LIST: {
+                    auto ret = rhs.getList();
+                    ret.values.insert(ret.values.begin(), lhs);
```
Shylock-Hg, line 1729: This may change the origin List value.

src/common/datatypes/test/ValueTest.cpp
```diff
@@ -89,6 +94,26 @@ TEST(Value, Arithmetics) {
         v = vStr1 + vDate1;
         EXPECT_EQ(Value::Type::STRING, v.type());
         EXPECT_EQ(std::string("Hello 2020-01-01"), v.getStr());
+
+        v = vList1 + vList2;
+        EXPECT_EQ(Value::Type::LIST, v.type());
+        EXPECT_EQ(List({1, 3, 2, 6, 4, 5}), v.getList());
+
+        v = vList1 + vBool2;
```
Shylock-Hg, line 102: Please check the origin List value.

jievince: check which list?

src/common/datatypes/Value.cpp
jievince: Why? It changed the **copy** of the original list.

Shylock-Hg: Yes, my mistake.

Shylock-Hg (approved):

PR #365, opened 2020-12-25
ignore suggest-override warning in gcc

critical27:
In https://github.com/vesoft-inc/nebula-storage/pull/232 use gmock, however, for now our gtest version is outdated, so we could only use the old style macro, which can not mark override.

We can ignore it by adding `#pragma GCC diagnostic ignored "-Wsuggest-override"`, but we failed to compile in clang for now, this would raise clang `unknown-warning-option` warning.

If we try to `#pragma clang diagnostic ignored "-Wunknown-warning-option"`, gcc can't ignore it anymore.

For similarity, just ignore `suggest-override` in gcc, and we could add it back when third-party update.


dutor (commented):
Better add a TODO comment here. Since we really benefit from this option.

critical27:
> Better add a TODO comment here. Since we really benefit from this option.

👌

dutor (approved):
LGTM.

bright-starry-sky (approved):
LGTM

darionyaphet (approved):

PR #371, opened 2020-12-29
Fix parallel thrift codegen

dutor:
This PR fixes two issues:
 * Thrift codegen is performed multiple times in parallel build.
 * Storage related sources might be built twice even nothing changed.

The reasons are:
 * `add_custom_command` iteself does not introduce a target.
 * Multiple targets depend on the generated files, i.e. the generating procedure, while these targets don't depend on each other. So the codegen will be invoked in each depending chain.
 * storage.thrift depends on meta.thrift just like common.thrift, but the dependency is not addressed explicitly.


yixinglu (approved):

bright-starry-sky (approved):
💯

PR #377, opened 2021-01-04
Add flag to determine whether partial success results are valid

Aiee:
As title.

relevant pr: https://github.com/vesoft-inc/nebula-graph/pull/556


yixinglu (approved):

yixinglu:
Don't forget to update clients libraries.

laura-ding:
> Don't forget to update clients libraries.

here is not modify the thrift file, why need update clients.
@Aiee do you need to add a type in graph.thrift?

yixinglu (approved):

yixinglu:
> > Don't forget to update clients libraries.
> 
> here is not modify the thrift file, why need update clients.
> @Aiee do you need to add a type in graph.thrift?

My fault. LGTM

jievince (approved):

PR #379, opened 2021-01-06
Add getAllLatestVerTagSchema()

Aiee:
Add getAllLatestTagSchema() which returns the latest version of schemas of all tags in the given space.

related PRs:
https://github.com/vesoft-inc/nebula-storage/pull/288
https://github.com/vesoft-inc/nebula-graph/pull/553

### DIFF
--- a/src/common/clients/meta/MetaClient.cpp
+++ b/src/common/clients/meta/MetaClient.cpp
@@ -1958,7 +1958,7 @@ StatusOr<TagSchemas> MetaClient::getAllVerTagSchema(GraphSpaceID spaceId) {
 }
 
 
-StatusOr<TagSchemas> MetaClient::getAllLatestTagSchema(const GraphSpaceID& spaceId) {
+StatusOr<TagLatestSchema> MetaClient::getAllLatestTagSchema(const GraphSpaceID& spaceId) {
     if (!ready_) {
         return Status::Error("Not ready!");
     }
@@ -1968,12 +1968,11 @@ StatusOr<TagSchemas> MetaClient::getAllLatestTagSchema(const GraphSpaceID& space
         return Status::Error("Space not %d found", spaceId);
     }
     std::unordered_map<TagID,
-        std::vector<std::shared_ptr<const NebulaSchemaProvider>>> tagsSchema;
+        std::shared_ptr<const NebulaSchemaProvider>> tagsSchema;
+    tagsSchema.reserve(iter->second->tagSchemas_.size());
     // fetch all tagIds
     for (const auto& tagSchema : iter->second->tagSchemas_) {
-        std::vector<std::shared_ptr<const NebulaSchemaProvider>> tagLatestSchema;
-        tagLatestSchema.emplace_back(tagSchema.second.back());
-        tagsSchema.emplace(tagSchema.first, tagLatestSchema);
+        tagsSchema.emplace(tagSchema.first, tagSchema.second.back());
     }
     return tagsSchema;
 }
--- a/src/common/clients/meta/MetaClient.cpp
+++ b/src/common/clients/meta/MetaClient.cpp
@@ -1952,23 +1952,22 @@ StatusOr<TagSchemas> MetaClient::getAllVerTagSchema(GraphSpaceID spaceId) {
     folly::RWSpinLock::ReadHolder holder(localCacheLock_);
     auto iter = localCache_.find(spaceId);
     if (iter == localCache_.end()) {
-        return Status::Error("Space not %d found", spaceId);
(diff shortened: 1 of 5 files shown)


src/common/clients/meta/MetaClient.cpp
```diff
@@ -1956,6 +1956,27 @@ StatusOr<TagSchemas> MetaClient::getAllVerTagSchema(GraphSpaceID spaceId) {
...
+    if (!ready_) {
+        return Status::Error("Not ready!");
+    }
+    folly::RWSpinLock::ReadHolder holder(localCacheLock_);
+    auto iter = localCache_.find(spaceId);
+    if (iter == localCache_.end()) {
+        return Status::Error("Space not %d found", spaceId);
+    }
+    std::unordered_map<TagID,
+        std::vector<std::shared_ptr<const NebulaSchemaProvider>>> tagsSchema;
```
Shylock-Hg, line 1969: reserve

Shylock-Hg, line 1969: Why need vector?

Aiee: Since currently the code uses vectors, it would be smoother to replace old interfaces with the same datatype, all we need to do is to change the function name.

Shylock-Hg: I don't think so.

Aiee: I'll modify this in the next commit

Shylock-Hg: ok

src/common/meta/SchemaManager.h
```diff
@@ -20,6 +20,10 @@ class MetaClient;
 using TagSchemas =
     std::unordered_map<TagID, std::vector<std::shared_ptr<const NebulaSchemaProvider>>>;
 
+// Mapping of tags and corresponding *latest* tag schema
+using TagLatestSchema =
```
Shylock-Hg, line 24: TagSchema is enough.

Shylock-Hg: I means rename the `TagLatestSchema` to `TagSchema` is better.

src/common/clients/meta/MetaClient.cpp
```diff
@@ -1956,6 +1956,26 @@ StatusOr<TagSchemas> MetaClient::getAllVerTagSchema(GraphSpaceID spaceId) {
...
 
 
+StatusOr<TagLatestSchema> MetaClient::getAllLatestTagSchema(const GraphSpaceID& spaceId) {
+    if (!ready_) {
+        return Status::Error("Not ready!");
+    }
+    folly::RWSpinLock::ReadHolder holder(localCacheLock_);
+    auto iter = localCache_.find(spaceId);
+    if (iter == localCache_.end()) {
+        return Status::Error("Space not %d found", spaceId);
```
yixinglu, line 1966:
```
"Space `%s' not found"
```
It's better to raise the space name to user in error message.

src/common/clients/meta/MetaClient.cpp
```diff
@@ -1956,6 +1956,26 @@ StatusOr<TagSchemas> MetaClient::getAllVerTagSchema(GraphSpaceID spaceId) {
...
+    }
+    std::unordered_map<TagID,
+        std::shared_ptr<const NebulaSchemaProvider>> tagsSchema;
```
yixinglu, line 1969: `TagLatestSchema tagsSchema;`

src/common/meta/SchemaManager.h
yixinglu: Why we need to declare same type definition in different header files ? Why not use the same declaration?

Aiee: According to my observation, the only case we use the mapping of tagId and its schema is when we need a pair of tagId and corresponding latest schema, and I think this naming explicitly indicates this.

Shylock-Hg: The type has not constraint to `latest`, so better not named so.

Aiee: Ok it makes sense. I'll rename it in the next PR.

src/common/clients/meta/MetaClient.cpp
Aiee: Since the input of this function is a spaceId, if the Id doesn't exist then there is no corresponding space name.

Shylock-Hg (approved):

Shylock-Hg (approved):

yixinglu (approved):

PR #380, opened 2021-01-06
support gets the value at specified row, col of a dataset

jievince:


nevermore3 (approved):

src/common/function/FunctionManager.cpp
```diff
@@ -167,8 +167,6 @@ std::unordered_map<std::string, std::vector<TypeSignature>> FunctionManager::typ
               TypeSignature({Value::Type::MAP}, Value::Type::DATETIME)}},
     {"timestamp", {TypeSignature({Value::Type::STRING}, Value::Type::INT),
                    TypeSignature({Value::Type::INT}, Value::Type::INT)}},
-    {"id", {TypeSignature({Value::Type::VERTEX}, Value::Type::STRING),
-             }},
```
yixinglu, line 171: Why remove it ?

jievince: @nevermore3  let it be removed from typeSignature_.

jievince: typeSignature_ is just used here:
```
void DeduceTypeVisitor::visit(FunctionCallExpression *expr) {
    std::vector<Value::Type> argsTypeList;
    argsTypeList.reserve(expr->args()->numArgs());
    for (auto &arg : expr->args()->args()) {
        arg->accept(this);
        if (!ok()) return;
        argsTypeList.push_back(type_);
    }
    auto funName = *expr->name();
    if (funName == "id" || funName == "src" || funName == "dst") {
        type_ = vidType_;
        return;
    }
    auto result = FunctionManager::getReturnType(funName, argsTypeList);
    if (!result.ok()) {
        status_ = Status::SemanticError("`%s` is not a valid expression : %s",
                                        expr->toString().c_str(),
                                        result.status().toString().c_str());
        return;
    }
    type_ = result.value();
}
```
And removing it from typeSignature_ doesn't  affect calling it

yixinglu (approved):

PR #386, opened 2021-01-13
add one error code

liuyu85cn:


src/common/interface/meta.thrift
```diff
@@ -76,6 +76,7 @@ enum ErrorCode {
     E_SAVE_JOB_FAILURE       = -57,
     E_BALANCER_FAILURE       = -58,
     E_JOB_NOT_FINISHED       = -59,
+    E_TASK_REPORT_OUT_DATE   = -80,
```
panda-sheep, line 79: `-80`， Why is it here?
Best to put the error codes in order?

liuyu85cn: because this is an admin error code, and I don't want split admin error code.

panda-sheep (approved):
Good，LGTM！

bright-starry-sky (approved):
LGTM

PR #389, opened 2021-01-14
Correct/ast to string format

Shylock-Hg:


Aiee (dismissed):

src/common/expression/AggregateExpression.cpp
```diff
@@ -64,7 +64,18 @@ const Value& AggregateExpression::eval(ExpressionContext& ctx) {
...
+    if (arg_->kind() == Expression::Kind::kConstant) {
+        const auto *argExpr = static_cast<const ConstantExpression*>(arg_.get());
+        if (argExpr->value().isStr()) {
+            arg = argExpr->value().getStr();
+        } else {
+            arg = arg_->toString();
+        }
+    } else {
+        arg = arg_->toString();
+    }
```
yixinglu, line 78: Why not to fix ConstantExpression::toString and let it return `value_->toString()` ? And then don't need to change here and following file.

Shylock-Hg: Emm in fact the ConstantExpression::toString implemented by value_->toString already.

Shylock-Hg: This is for the agg such as `count(*)` the `*` will stored as ConstantExpression(string), so need a trick for it.

jievince (approved):

src/common/expression/AggregateExpression.cpp
yixinglu: But according to your implementation,  the format of `count(*)` expression will seems to be the string `count("*")`. Is it expected?  could you add some test cases for this scenario?

Shylock-Hg: No, it's `count(*)` in my PR.The case cover it.

yixinglu (dismissed):

nevermore3 (approved):

CPWstatic (approved):

PR #390, opened 2021-01-17
Fix path duplicate functions and add value hash/equal benchmark

yixinglu:
benchmark between string and value.

### DIFF
--- a/src/common/datatypes/Path.cpp
+++ b/src/common/datatypes/Path.cpp
@@ -8,8 +8,10 @@
 
 #include <folly/hash/Hash.h>
 #include <folly/String.h>
+#include <tuple>
 #include <unordered_set>
 #include "common/datatypes/Path.h"
+#include "common/thrift/ThriftTypes.h"
 
 namespace nebula {
 void Path::reverse() {
@@ -40,11 +42,10 @@ bool Path::hasDuplicateVertices() const {
     if (steps.empty()) {
         return false;
     }
-    std::unordered_set<Value> uniqueVid;
-    auto srcVid = src.vid;
-    uniqueVid.emplace(srcVid);
+    std::unordered_set<std::string> uniqueVid;
+    uniqueVid.emplace(src.vid.toString());
     for (const auto& step : steps) {
-        auto ret = uniqueVid.emplace(step.dst.vid);
+        auto ret = uniqueVid.emplace(step.dst.vid.toString());
         if (!ret.second) {
             return true;
         }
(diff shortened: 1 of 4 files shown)


nevermore3 (dismissed):

nevermore3 (approved):

src/common/datatypes/Path.cpp
```diff
@@ -57,20 +57,24 @@ bool Path::hasDuplicateEdges() const {
...
+        const auto& edgeSrc = step.type > 0 ? srcVid : dstVid;
+        const auto& edgeDst = step.type > 0 ? dstVid : srcVid;
+        auto edgeKey = folly::stringPrintf("%s%ld%s%ld%s%ld%ld",
+                                           edgeSrc.c_str(),
+                                           edgeSrc.size(),
+                                           edgeDst.c_str(),
+                                           edgeDst.size(),
                                            step.name.c_str(),
+                                           step.name.size(),
                                            step.ranking);
```
CPWstatic, line 72: Would it be faster by using tuple?

yixinglu: Yeah. Fixed.

src/common/datatypes/Path.cpp
```diff
@@ -56,21 +57,18 @@ bool Path::hasDuplicateEdges() const {
     if (steps.size() < 2) {
         return false;
     }
-    std::unordered_set<std::string> uniqueSet;
+    using Key = std::tuple<Value, Value, std::string, EdgeRanking>;
```
Shylock-Hg, line 60: I think you could use edgeType instead of edge name.

yixinglu: OK, fixed.

Aiee (approved):

src/common/datatypes/Path.cpp
```diff
@@ -56,21 +57,18 @@ bool Path::hasDuplicateEdges() const {
...
-        auto edgeDst = step.type > 0 ? step.dst.vid : srcVid;
-        auto edgeKey = folly::stringPrintf("%s%s%s%ld",
-                                           edgeSrc.toString().c_str(),
-                                           edgeDst.toString().c_str(),
-                                           step.name.c_str(),
-                                           step.ranking);
-        auto res = uniqueSet.emplace(std::move(edgeKey));
-        if (!res.second) {
+        const auto& dstVid = step.dst.vid;
+        const auto& edgeSrc = step.type > 0 ? srcVid : dstVid;
```
laura-ding, line 65: It's not needed here anymore

laura-ding (approved):

PR #392, opened 2021-01-20
Enhanced StatsManager utility

sherman-the-tank:
The major change is to make readAllValue() only return the matrix that specified at the registeration time


src/common/stats/StatsManager.cpp
```diff
@@ -89,10 +143,17 @@ int32_t StatsManager::registerHisto(folly::StringPiece counterName,
                 max,
                 StatsType(60, {seconds(5), seconds(60), seconds(600), seconds(3600)}))));
```
Shylock-Hg, line 144: Do we need to config the time range?

sherman-the-tank: It's better to pre-set the range, otherwise it's going to be a mess

sherman-the-tank: We can say we only provide these 4 ranges, which I think is more than enough

Shylock-Hg (approved):

src/common/stats/StatsManager.cpp
```diff
@@ -184,93 +263,101 @@ void StatsManager::readAllValue(folly::dynamic& vals) {
...
+               case StatsMethod::RATE:
+                    metricPrefix += ".rate";
+                    break;
+                // intentionally no `default'
+            }
 
-                switch (method) {
-                    case StatsMethod::SUM:
-                        metricName += ".sum";
+            for (auto range = TimeRange::FIVE_SECONDS;
```
critical27, line 285: Maybe we could add a parameter to specify which time range. Since we have registered a lot stats already, print all of them will be messy, and usually user would exactly know which time range they need. Of course, by default we could return all value with all time range.

sherman-the-tank: This method is mainly for external monitoring system to collect the counters. So if we don't collect at this time, we will never be able to see the counters for the specific time period in the external monitoring system

sherman-the-tank: Although, I think 10sec period might be more useful than 5sec period

yixinglu (approved):

Shylock-Hg (approved):

critical27:
Storage PR is ready. https://github.com/vesoft-inc/nebula-storage/pull/342

PR #393, opened 2021-01-20
storage client no retry， retry is completely invalid.

panda-sheep:
The current storage client retry mechanism is completely wrong.
For example
1）a host's request contains 2 parts, a part succeeds, the other has leader change,
2）a host's request contains 3 parts, one part is successful, and the other two parts have leader change, and these two parts correspond to two leaders respectivelys.
In two way, how many times the request is retried is always an error. Retrying is completely invalid.

There are two solutions:
1) The storage client does not retry, modify the leader host of all the parts of the leader change, so the next time you re-execute sql, it will succeed
2) Storage client retry. When the errcode in the same request is leader change, it will retry, otherwise it will not retry. Need to reorganize the leader change requst in the same request of collectResponse and getResponse. Then it may be sent multiple times.

The current pr is not to retry for storage client，I prefer storage client not to retry
Please hold this pr first for chaos test

The retry of storage client  pr will also perform the chaos test.

### DIFF
--- a/src/common/clients/storage/InternalStorageClient.cpp
+++ b/src/common/clients/storage/InternalStorageClient.cpp
@@ -84,8 +84,7 @@ void InternalStorageClient::forwardTransactionImpl(int64_t txnId,
         std::make_pair(dest, interReq),
         [](cpp2::InternalStorageServiceAsyncClient* client, const cpp2::InternalTxnRequest& r) {
             return client->future_forwardTransaction(r);
-        },
-        kInternalPortOffset)
+        })
         .thenTry([=, p = std::move(p)](auto&& t) mutable {
             auto code = extractErrorCode(t);
             if (code == cpp2::ErrorCode::E_LEADER_CHANGED) {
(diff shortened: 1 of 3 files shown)


Shylock-Hg:
Good catch. Maybe we could reorganize the request of  one host as `map<PartID, Request>` in this form it's easy to reform when retry request.

panda-sheep:
> map
shoud map<leaderHost, Request> Both ways are ok, I am testing

Shylock-Hg:
> > map
> > shoud map<leaderHost, Request>
> > Both ways are ok, I am testing

I means map<Host, map<PartID, Request>>

panda-sheep:
> > > map
> > > shoud map<leaderHost, Request>
> > > Both ways are ok, I am testing
> 
> I means map<Host, map<PartID, Request>>

Shouldn't be like this,
The `request `contains the parts contained in the request,  and the redundant part contained in the `request   `should be removed

critical27 (approved):
LGTM

CPWstatic (approved):

PR #395, opened 2021-01-22
Fix the retry partition conflict.

Shylock-Hg:
Fix the issue in #393 , that now will collect the failed partition and reorganize the request and retry. The promise will set when no partition retry or exceeded retryLimit.


Shylock-Hg:
Stale

PR #396, opened 2021-01-22
fix agg and List operator

czpmango:
Fix [https://github.com/vesoft-inc/nebula-graph/issues/678](https://github.com/vesoft-inc/nebula-graph/issues/678)


yixinglu (approved):
LGTM

jievince (approved):

PR #401, opened 2021-01-28
fix RelationalExpression

czpmango:
This PR fixes some bugs about Value comparison related to `null` and `empty`.  Fixes https://github.com/vesoft-inc/nebula-graph/issues/644
There are some test cases:
 - 2>true : null   2<=true : null
 - 2==true : false
 - null==true : null
 - null<>null : null
 - empty==empty : true
 - empty==null : null
 - empty>=1 : empty

### DIFF
--- a/src/common/datatypes/Value.cpp
+++ b/src/common/datatypes/Value.cpp
@@ -1629,6 +1629,182 @@ Value Value::toInt() {
     }
 }
 
+Value Value::lessThan(const Value& v) const {
+    if (empty() || v.empty()) {
+       return (v.isNull() || isNull())? Value::kNullValue : Value::kEmpty;
+    }
+    auto vType = v.type();
+    auto hasNull = (type_ | vType) & Value::Type::NULLVALUE;
+    auto notSameType = type_ != vType;
+    auto notBothNumeric = ((type_ | vType) & Value::kNumericType) != Value::kNumericType;
+    if (hasNull || (notSameType && notBothNumeric)) {
+        return Value::kNullValue;
+    }
+
+    switch (type_) {
+        case Value::Type::BOOL: {
+            return getBool() < v.getBool();
+        }
+        case Value::Type::INT: {
+            switch (vType) {
+                case Value::Type::INT: {
+                    return getInt() < v.getInt();
+                }
+                case Value::Type::FLOAT: {
+                    return (std::abs(getInt() - v.getFloat()) >= kEpsilon
+                            && getInt() < v.getFloat());
+                }
+                default: {
+                    return kNullBadType;
+                }
+            }
+        }
+        case Value::Type::FLOAT: {
+            switch (vType) {
+                case Value::Type::INT: {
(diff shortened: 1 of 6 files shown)


yixinglu (commented):
It's not a good practice that fixing bug and refactoring in one PR since i don't clearly check where you have changed.

czpmango:
> It's not a good practice that fixing bug and refactoring in one PR since i don't clearly check where you have changed.

No changes have been made to the behavior of the class `Value`. Just fix RelationalExpression’s behavior, and this requires adding interfaces `Value::lessThan` to allow errors to be passed to the Expression layer.

yixinglu (approved):

nevermore3 (dismissed):

src/common/datatypes/Value.cpp
```diff
@@ -2226,164 +2402,22 @@ Value operator!(const Value& rhs) {
...
-            return false;
-        }
-        case Value::Type::NULLVALUE:
-        case Value::Type::__EMPTY__: {
-            return false;
-        }
+    auto res = lhs.lessThan(rhs);
+    if (res.isBool()) {
+        return res.getBool();
+    } else {
```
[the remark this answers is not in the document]
czpmango: Already removed.

yixinglu (dismissed):

nevermore3 (approved):

yixinglu (approved):

jievince (approved):

src/common/expression/RelationalExpression.cpp
```diff
@@ -15,31 +15,24 @@ const Value& RelationalExpression::eval(ExpressionContext& ctx) {
...
-            result_ = lhs < rhs;
+            result_ = lhs.lessThan(rhs);
             break;
         case Kind::kRelLE:
-            result_ = lhs <= rhs;
+            result_ = lhs.lessThan(rhs) || lhs.equal(rhs);
             break;
         case Kind::kRelGT:
-            result_ = lhs > rhs;
+            result_ = !lhs.lessThan(rhs) && !lhs.equal(rhs);
```
Shylock-Hg, line 32: What's the result of `!NULL`?

src/common/expression/RelationalExpression.cpp
```diff
@@ -15,31 +15,24 @@ const Value& RelationalExpression::eval(ExpressionContext& ctx) {
...
-            result_ = lhs != rhs;
+            result_ = !lhs.equal(rhs);
             break;
         case Kind::kRelLT:
-            result_ = lhs < rhs;
+            result_ = lhs.lessThan(rhs);
             break;
         case Kind::kRelLE:
-            result_ = lhs <= rhs;
+            result_ = lhs.lessThan(rhs) || lhs.equal(rhs);
```
Shylock-Hg, line 29: And why not call <= directly.

nevermore3 (approved):

src/common/expression/RelationalExpression.cpp
czpmango: add static method?

PR #405, opened 2021-02-01
storage client retry

panda-sheep:
As title

### DIFF
--- a/src/common/clients/storage/StorageClientBase.h
+++ b/src/common/clients/storage/StorageClientBase.h
@@ -34,6 +34,10 @@ public:
 
     explicit StorageRpcResponse(size_t reqsSent) : totalReqsSent_(reqsSent) {}
 
+    void increase(size_t size) {
+        totalReqsSent_ += size;
+    }
+
     bool succeeded() const {
         return result_ == Result::ALL_SUCCEEDED;
     }
@@ -85,7 +89,7 @@ public:
     }
 
 private:
-    const size_t totalReqsSent_;
+    size_t totalReqsSent_;
     size_t failedReqs_{0};
 
     Result result_{Result::ALL_SUCCEEDED};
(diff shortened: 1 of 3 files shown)


src/common/clients/storage/StorageClientBase.h
```diff
@@ -196,11 +268,44 @@ class StorageClientBase {
         return partsId;
     }
 
+    template <typename K>
+    std::unordered_map<PartitionID, K> getReqPartsIdFromContainer(
+        const std::unordered_map<PartitionID, K> &t, PartitionID partId) const {
+        std::unordered_map<PartitionID, K> ret;
+        ret.emplace(partId, t.at(partId));
```
monadbobo, line 275: Perhaps it is more straightforward to do so.

`return std::unordered_map<PartitionID, K>{{partId, t.at(partId)}};`

src/common/clients/storage/StorageClientBase.h
```diff
@@ -196,11 +268,44 @@ class StorageClientBase {
...
+        return ret;
+    }
+
+    template <typename K>
+    void addPartRequest(std::unordered_map<PartitionID, K> &oldMap,
```
monadbobo, line 280: `const std::unordered_map<PartitionID, K> &oldMap`

src/common/clients/storage/StorageClientBase.h
```diff
@@ -196,11 +268,44 @@ class StorageClientBase {
...
+                        std::unordered_map<PartitionID, K> &newMap,
+                        PartitionID partId) {
+        auto it = newMap.find(partId);
+        if (it != newMap.end()) {
+            it->second.insert(newMap.at(partId).end(),
+                              oldMap.at(partId).begin(), oldMap.at(partId).end());
```
monadbobo, line 286: It might be better to use the move iterator(oldMap) here.

src/common/clients/storage/StorageClientBase.h
panda-sheep: Good，thx

PR #406, opened 2021-02-01
Handle the responses via different evb.

CPWstatic:


nevermore3 (approved):

src/common/clients/storage/StorageClientBase.inl
```diff
@@ -176,16 +176,14 @@ StorageClientBase<ClientType>::collectResponse(
...
-        evb = ioThreadPool_->getEventBase();
-    }
+    DCHECK(!!ioThreadPool_);
 
     for (auto& req : requests) {
         auto& host = req.first;
         auto spaceId = req.second.get_space_id();
         auto res = context->insertRequest(host, std::move(req.second));
         DCHECK(res.second);
+        evb = ioThreadPool_->getEventBase();
```
Shylock-Hg, line 186: I think you could call it outside the loop.

src/common/clients/storage/StorageClientBase.h
```diff
@@ -32,59 +32,77 @@ class StorageRpcResponse final {
...
         return responses_;
     }
 
+    // Not thread-safe.
     const std::vector<std::tuple<HostAddr, int32_t, int32_t>>& hostLatency() const {
         return hostLatency_;
     }
 
 private:
+    std::shared_ptr<std::mutex> lock_;
```
monadbobo, line 105: Why is shared_ptr\<std::mutex\> used here instead of using std::mutex directly?

CPWstatic: We had some compiling problems if we use std::mutex directly, since the std::mutex is uncopyable.

src/common/clients/storage/StorageClientBase.h
```diff
@@ -32,59 +32,77 @@ class StorageRpcResponse final {
...
+        std::lock_guard<std::mutex> g(*lock_);
+        responses_.emplace_back(std::move(resp));
+    }
+
+    // Not thread-safe.
+    const std::unordered_map<PartitionID, storage::cpp2::ErrorCode>& failedParts() const {
+        return failedParts_;
+    }
+
+    // Not thread-safe.
```
laura-ding, line 94: why not make it thread-safe

CPWstatic: These interfaces return reference, and they were used after all request are collected.

dutor (approved):
LGTM.

nevermore3 (approved):

PR #407, opened 2021-02-02
Cache the function when function call evaluation to avoid duplicate l…

Shylock-Hg:
…ookup.

We don't provide method to modify the function name so the cache is correct always.


src/common/expression/FunctionCallExpression.cpp
```diff
@@ -69,15 +69,20 @@ void FunctionCallExpression::resetFrom(Decoder& decoder) {
 }
 
 const Value& FunctionCallExpression::eval(ExpressionContext& ctx) {
-    auto function = FunctionManager::get(*name_, args_->numArgs());
-    if (!function.ok()) {
-        result_ = Value::kNullBadData;
-    } else {
+    if (!func_.hasValue()) {
```
yixinglu, line 72: Why not to initialize the function in FunctionCallExpression constructor? When we construct the FunctionCallExpression, we have known which function to call.

Shylock-Hg: The decode will create a empty FunctionCallExpression.

jievince (approved):

Shylock-Hg:
Benchmark:
before this PR:
============================================================================
/home/shylock.huang/nebula-common2/src/common/expression/test/FunctionCallExpressionBenchmark.cpprelative  time/iter  iters/s
============================================================================
funcCall(FunctionCallBM)                                   807.59ns    1.24M
============================================================================
after this PR:
============================================================================
/home/shylock.huang/nebula-common/src/common/expression/test/FunctionCallExpressionBenchmark.cpprelative  time/iter  iters/s
============================================================================
funcCall(FunctionCallBM)                                   373.90ns    2.67M
============================================================================

jievince (dismissed):

src/common/expression/FunctionCallExpression.cpp
yixinglu: I don't understand what you mean by the relationship between function cache and decoder? I agree the optimization but it's better to do this in construction of FunctionCallExpression rather than `eval` function.

Shylock-Hg: Fixed.

czpmango (commented):
The encode/decode function should be modified accordingly?  I wanted to do something similar to AggregateExpression before, but I got stuck on encode this lambda.

Shylock-Hg:
> The encode/decode function should be modified accordingly? I wanted to do something similar to AggregateExpression before, but I got stuck on encode this lambda.

I think don't need to encode the lambda, just encode the function name and lookup once when decode.

src/common/expression/FunctionCallExpression.cpp
```diff
@@ -66,18 +66,22 @@ void FunctionCallExpression::resetFrom(Decoder& decoder) {
...
-    } else {
+    if (func_.hasValue()) {
         std::vector<Value> parameter;
         for (const auto& arg : args_->args()) {
             parameter.emplace_back(std::move(arg->eval(ctx)));
         }
-        result_ = function.value()(parameter);
+        result_ = func_.value()(parameter);
+    } else {
+        result_ = Value::kNullBadData;
```
czpmango, line 84: If cache missed, don't need to call function  `FunctionManager::get` again?

Shylock-Hg: Don't need it's done in constructor and decode.

czpmango (dismissed):
Good Job.

jievince (dismissed):
LGTM

src/common/expression/test/FunctionCallExpressionBenchmark.cpp
```diff
@@ -0,0 +1,40 @@
+/* Copyright (c) 2020 vesoft inc. All rights reserved.
```
darionyaphet, line 1: 2021

Shylock-Hg: Fixed.

czpmango (dismissed):

nevermore3 (dismissed):

src/common/expression/FunctionCallExpression.cpp
```diff
@@ -66,18 +66,22 @@ void FunctionCallExpression::resetFrom(Decoder& decoder) {
...
+        func_ = std::move(funcResult).value();
+    }
 }
 
 const Value& FunctionCallExpression::eval(ExpressionContext& ctx) {
-    auto function = FunctionManager::get(*name_, args_->numArgs());
-    if (!function.ok()) {
-        result_ = Value::kNullBadData;
-    } else {
+    if (func_.hasValue()) {
```
yixinglu, line 77: It's not necessary to check whether this function is valid. It should be always not nullable value, it's better to replace this check with DCHECK and the optional is also necessary.

Shylock-Hg: This is to keep same with before that the invalid function call return null value.

Shylock-Hg: Fixed.

yixinglu (approved):

nevermore3 (approved):

PR #408, opened 2021-02-02
fix null toString

czpmango:
Fixes [https://github.com/vesoft-inc/nebula-graph/issues/720](https://github.com/vesoft-inc/nebula-graph/issues/720)


src/common/function/FunctionManager.cpp
```diff
@@ -798,7 +798,7 @@ FunctionManager::FunctionManager() {
         attr.body_ = [](const auto &args) -> Value {
              switch (args[0].type()) {
```
Shylock-Hg, line 799: Why don't call Value::toString

czpmango: Be consistent with Cypher.

jievince (approved):

yixinglu (approved):

PR #409, opened 2021-02-03
Add isNull/isNotNull operator

czpmango:
Add `IS NULL` and `IS NOT NULL`

Fixes [https://github.com/vesoft-inc/nebula-graph/issues/701](https://github.com/vesoft-inc/nebula-graph/issues/701)

### DIFF
--- a/src/common/expression/UnaryExpression.cpp
+++ b/src/common/expression/UnaryExpression.cpp
@@ -39,7 +39,8 @@ void UnaryExpression::resetFrom(Decoder& decoder) {
 
 
 const Value& UnaryExpression::eval(ExpressionContext& ctx) {
-   switch (kind_) {
+    DCHECK(!!operand_);
+    switch (kind_) {
         case Kind::kUnaryPlus: {
             Value val(operand_->eval(ctx));
             result_ = std::move(val);
@@ -75,6 +76,14 @@ const Value& UnaryExpression::eval(ExpressionContext& ctx) {
             ctx.setVar(varExpr->var(), result_);
             break;
         }
+        case Kind::kIsNull: {
+            result_ = (operand_->eval(ctx)).isNull() ? true : false;
+            break;
+        }
+        case Kind::kIsNotNull: {
+            result_ = (operand_->eval(ctx)).isNull() ? false : true;
+            break;
+        }
        default:
            LOG(FATAL) << "Unknown type: " << kind_;
    }
@@ -98,6 +107,12 @@ std::string UnaryExpression::toString() const {
             break;
         case Kind::kUnaryDecr:
             op = "--";
+            break;
+        case Kind::kIsNull:
+            op = "IS NULL";
+            break;
+        case Kind::kIsNotNull:
+            op = "IS NOT NULL";
             break;
         default:
             op = "illegal symbol ";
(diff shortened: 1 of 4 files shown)


src/common/expression/UnaryExpression.cpp
```diff
@@ -75,6 +76,14 @@ const Value& UnaryExpression::eval(ExpressionContext& ctx) {
...
             break;
         }
+        case Kind::kIsNull: {
+            result_ = (operand_->eval(ctx)).isNull() ? true : false;
+            break;
+        }
+        case Kind::kIsNotNull: {
+            result_ = (operand_->eval(ctx)).isNull() ? false : true;
+            break;
+        }
```
CPWstatic, line 86: These two ops should be kind of relational expression.

jievince (approved):

src/common/expression/UnaryExpression.cpp
```diff
@@ -39,6 +39,7 @@ void UnaryExpression::resetFrom(Decoder& decoder) {
 
 
 const Value& UnaryExpression::eval(ExpressionContext& ctx) {
+    DCHECK(!!operand_);
    switch (kind_) {
```
laura-ding, line 43: Adjust the indentation of the `switch`

src/common/expression/UnaryExpression.cpp
laura-ding: agree with @CPWstatic 
convert to `expression == NULL` or `expression != NULL`

laura-ding: agree with @CPWstatic 
convert to `expression == NULL` or `expression != NULL`

czpmango: > agree with @CPWstatic
> convert to `expression == NULL` or `expression != NULL`

For openCypher `expression == null` will always get null, should use `isNull` to judge null equality.

src/common/expression/UnaryExpression.cpp
czpmango: My mistake, thx~

src/common/expression/UnaryExpression.cpp
```diff
@@ -99,6 +108,12 @@ std::string UnaryExpression::toString() const {
         case Kind::kUnaryDecr:
             op = "--";
             break;
+        case Kind::kIsNull:
+            op = "IS NULL";
```
laura-ding, line 112: here maybe is " IS NULL", miss a space. you can add test for it

src/common/expression/UnaryExpression.cpp
```diff
@@ -99,6 +108,12 @@ std::string UnaryExpression::toString() const {
...
+            break;
+        case Kind::kIsNotNull:
+            op = "IS NOT NULL";
```
laura-ding, line 115: ditto

src/common/expression/UnaryExpression.cpp
czpmango: Already add test case.  
`2 IS NULL` or `IS NULL(2)`, Which way do you think is better?

laura-ding: Consistent with the grammar is ok

czpmango: ok.

Aiee (approved):

jievince (approved):

PR #411, opened 2021-02-05
Fix problem: storage restarted but reconnect is not working.

guojun85:
Fix problem: when storaged restarted, reconnect is now working and still use the bad connection.

https://discuss.nebula-graph.com.cn/t/topic/2805/3

### DIFF
--- a/src/common/thrift/ThriftClientManager.inl
+++ b/src/common/thrift/ThriftClientManager.inl
@@ -27,7 +27,8 @@ std::shared_ptr<ClientType> ThriftClientManager<ClientType>::client(
     auto it = clientMap_->find(std::make_pair(host, evb));
     if (it != clientMap_->end()) {
         auto channel = dynamic_cast<apache::thrift::HeaderClientChannel*>(it->second->getChannel());
-        if (channel == nullptr || !channel->good() || channel->getTransport()->readable()) {
+        if (channel == nullptr || !channel->good()
+            || (evb->isInEventBaseThread() && channel->getTransport()->readable())) {
             // Remove bad connection to create a new one.
             clientMap_->erase(it);
             LOG(ERROR) << "Invalid Channel: " << channel << " for host: " << host;
@@ -74,7 +75,8 @@ std::shared_ptr<ClientType> ThriftClientManager<ClientType>::client(
         headerClientChannel->setProtocolId(apache::thrift::protocol::T_BINARY_PROTOCOL);
         headerClientChannel->setClientType(THRIFT_UNFRAMED_DEPRECATED);
     }
-    std::shared_ptr<ClientType> client(new ClientType(std::move(headerClientChannel)), [evb](auto* p) {
+    std::shared_ptr<ClientType> client(new ClientType(std::move(headerClientChannel)),
+            [evb](auto* p) {
         evb->runImmediatelyOrRunInEventBaseThreadAndWait([p] {
             delete p;
         });
--- a/src/common/thrift/ThriftClientManager.inl
+++ b/src/common/thrift/ThriftClientManager.inl
@@ -23,53 +23,58 @@ std::shared_ptr<ClientType> ThriftClientManager<ClientType>::client(
     if (evb == nullptr) {
         evb = folly::EventBaseManager::get()->getEventBase();
     }
-
+    // Get client from client manager if it is ok.
     auto it = clientMap_->find(std::make_pair(host, evb));
     if (it != clientMap_->end()) {
-        return it->second;
+        auto channel = dynamic_cast<apache::thrift::HeaderClientChannel*>(it->second->getChannel());
+        if (channel == nullptr || !channel->good() || channel->getTransport()->readable()) {
+            // Remove bad connection to create a new one.
+            clientMap_->erase(it);
+            LOG(ERROR) << "Invalid Channel: " << channel << " for host: " << host;
+        } else {
+            LOG(INFO) << "Get connection from client map, host: " << host;
+            return it->second;
+        }
     }
 
-    // Need to create a new client
+    // Need to create a new client and insert it to client map.
     VLOG(2) << "There is no existing client to " << host << ", trying to create one";
-    auto channel = apache::thrift::ReconnectingRequestChannel::newChannel(
-        *evb, [compatibility, host = host, timeout] (folly::EventBase& eb) mutable {
-            static thread_local int connectionCount = 0;
+    static thread_local int connectionCount = 0;
+    /*
+        * TODO(liuyu): folly said 'resolve' may take second to finish
+        *              if this really happen, we will add a cache here.
+        * */
+    HostAddr tempHost = host;
+    if (!folly::IPAddress::validate(tempHost.host)) {
+        try {
+            folly::SocketAddress socketAddr(tempHost.host, tempHost.port, true);
+            std::ostringstream oss;
+            oss << "resolve " << tempHost << " as ";
+            tempHost.host = socketAddr.getAddressStr();
+            oss << tempHost;
+            LOG(INFO) << oss.str();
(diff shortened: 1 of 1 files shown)


src/common/thrift/ThriftClientManager.inl
```diff
@@ -23,53 +23,58 @@ std::shared_ptr<ClientType> ThriftClientManager<ClientType>::client(
     if (evb == nullptr) {
         evb = folly::EventBaseManager::get()->getEventBase();
     }
-
+    // Get client from client manager if it is ok.
     auto it = clientMap_->find(std::make_pair(host, evb));
     if (it != clientMap_->end()) {
-        return it->second;
+        auto channel = dynamic_cast<apache::thrift::HeaderClientChannel*>(it->second->getChannel());
+        if (channel == nullptr || !channel->good() || channel->getTransport()->readable()) {
```
critical27, line 30: The `good()` has been checked in `ReconnectingRequestChannel::impl()`, it won't work for now? I think I tested it long before. Quite strange.

critical27 (commented):
Thx for contribution.

src/common/thrift/ThriftClientManager.inl
guojun85: good() only checks the state, if no read or write on the socket, socket state will not be changed.

It is very easy to reproduce this issue by restarting storaged, graphd will use the bad connection once, and got eof.

critical27: After restart storaged, I can't reproduce EOF, only `N6apache6thrift9transport19TTransportExceptionE: AsyncSocketException: connect failed, type = Socket not open`, and it do reconnect. Did you have any other operation?

guojun85: I didn't have other operations, did you have "connect failed, type = Socket not open" on storaged, I reproduced this problem on storaged. This still has problem, reconnect should be down before you use the connection, "connect failed, type = Socket not open"  occurs after you got the connection?

guojun85: It is very easy to reproduce this problem for only one machine, when you restart storaged and wait for a moment for the raft part leader election done. then then send a request to storage from graphd. you can find that the bad connection is used.

I0207 19:54:10.099781 22669 StorageClientBase.inl:126] No leader exists. Choose one random.
I0207 19:54:10.099884 22654 ThriftClientManager.inl:21] Getting a client to [127.0.0.1:9779]
I0207 19:54:10.099907 22654 ThriftClientManager.inl:35] Get connection from client map cache, host: [127.0.0.1:9779]
E0207 19:54:10.100073 22654 StorageClientBase.inl:213] Request to [127.0.0.1:9779] failed: N6apache6thrift9transport19TTransportExceptionE: Channel got EOF. Check for server hitting connection limit, server connection idle timeout, and server crashes.
I0207 19:54:10.100111 22654 StorageClientBase.inl:160] Invalidate the leader for [1, 6]
I0207 19:54:10.100181 22669 GetNeighborsExecutor.cpp:107] Get neighbors time: 437us
E0207 19:54:10.100221 22669 StorageAccessExecutor.h:34] GetNeighborsExecutor failed, error E_RPC_FAILURE, part 6
E0207 19:54:10.100242 22669 StorageAccessExecutor.h:109] Storage Error: part: 6, error: E_RPC_FAILURE(-3).
E0207 19:54:10.100308 22669 QueryInstance.cpp:134] Storage Error: part: 6, error: E_RPC_FAILURE(-3).

critical27: Ah, I got your point. After restart, the first request will `Channel got EOF`, only when here comes the second request, it will create a new channel.

src/common/thrift/ThriftClientManager.inl
```diff
@@ -23,53 +23,60 @@ std::shared_ptr<ClientType> ThriftClientManager<ClientType>::client(
...
+    // Get client from client manager if it is ok.
     auto it = clientMap_->find(std::make_pair(host, evb));
     if (it != clientMap_->end()) {
-        return it->second;
+        auto channel = dynamic_cast<apache::thrift::HeaderClientChannel*>(it->second->getChannel());
+        if (channel == nullptr || !channel->good()
+            || (evb->isInEventBaseThread() && channel->getTransport()->readable())) {
+            // Remove bad connection to create a new one.
+            clientMap_->erase(it);
+            LOG(ERROR) << "Invalid Channel: " << channel << " for host: " << host;
```
critical27, line 34: Would you lower the log level, VLOG(2) is enough I guess

src/common/thrift/ThriftClientManager.inl
```diff
@@ -23,53 +23,60 @@ std::shared_ptr<ClientType> ThriftClientManager<ClientType>::client(
     if (evb == nullptr) {
         evb = folly::EventBaseManager::get()->getEventBase();
     }
-
+    // Get client from client manager if it is ok.
     auto it = clientMap_->find(std::make_pair(host, evb));
     if (it != clientMap_->end()) {
-        return it->second;
+        auto channel = dynamic_cast<apache::thrift::HeaderClientChannel*>(it->second->getChannel());
```
critical27, line 29: Would you remove the duplicate `dynamic_cast` and `good()` in `ReconnectingRequestChannel::impl()`? After this PR merged, we don't need it anymore.

critical27 (commented):
Generally LGTM, a few lines need to be modified. Thx again.

src/common/thrift/ThriftClientManager.inl
guojun85: This fix still has some problem for storaged raft client. need to do more investigation.

src/common/thrift/ThriftClientManager.inl
```diff
@@ -23,53 +23,60 @@ std::shared_ptr<ClientType> ThriftClientManager<ClientType>::client(
...
+        if (channel == nullptr || !channel->good()
+            || (evb->isInEventBaseThread() && channel->getTransport()->readable())) {
```
panda-sheep, line 31: `channel->getTransport()->readable()`  return  true if the `transport `is readable, false otherwise.
Forgot to negate the operation here?

guojun85: No，if the client is not used but readable is true, means there is something wrong with current transport.

src/common/thrift/ThriftClientManager.inl
critical27: Any follow-ups? I have not test this PR recently.

src/common/thrift/ThriftClientManager.inl
panda-sheep: When `channel->good()` is true, and` channel->getTransport()->readable()` is false,  the channel should it be erased?

src/common/thrift/ThriftClientManager.inl
guojun85: It seems it will occasionally free good connection. I didn't find the root cause currently.

src/common/thrift/ThriftClientManager.inl
guojun85: No, When channel-good() is true, and channel->getTransport()->readable() is true. the channel should be erased. for the channel is not used but it is readable, which means something wrong with current channel. For example, server closed current socket, and it will be readable.

src/common/thrift/ThriftClientManager.inl
critical27: 👌 , I will dig more details later. As for master code, the reconnect do works, however it need to send one more request.

guojun85: Yes, I think this is an issue, and can be fixed.

critical27:
@guojun85 it has been fixed in #431, thx for the contribution.

PR #412, opened 2021-02-08
Fix TypeCastExpression

czpmango:
This PR fixes TypeCastExpression behavior(refer to: [openCypher scalar functions](https://neo4j.com/docs/cypher-manual/current/functions/scalar/#functions-toboolean )) and add functions related with type conversion.

Depended by [https://github.com/vesoft-inc/nebula-graph/pull/746](https://github.com/vesoft-inc/nebula-graph/pull/746) and [https://github.com/vesoft-inc/nebula-storage/pull/352](https://github.com/vesoft-inc/nebula-storage/pull/352)

Fixes [https://github.com/vesoft-inc/nebula-graph/issues/677](https://github.com/vesoft-inc/nebula-graph/issues/677) and [https://github.com/vesoft-inc/nebula-graph/issues/706](https://github.com/vesoft-inc/nebula-graph/issues/706)

---

Note:
```
Value::toBool() 行为有如下变动:
 - 除 null/bool/string 之外的类型视为非法类型，返回 kNullBadType
 - bool 类型原样返回
 - 字符串 "true"/"FaLse" 等会返回转换后的 bool 值
 - null 和其他字符串会返回 null
 - empty 类型行为暂未定义，目前返回 null
```

### DIFF
--- a/src/common/datatypes/Value.cpp
+++ b/src/common/datatypes/Value.cpp
@@ -1550,86 +1550,81 @@ std::string Value::toString() const {
     LOG(FATAL) << "Unknown value type " << static_cast<int>(type_);
 }
 
-std::pair<bool, bool> Value::toBool() {
+Value Value::toBool() {
     switch (type_) {
-        // Type::__EMPTY__ is always false
-        case Value::Type::__EMPTY__: {
-            return std::make_pair(false, true);
-        }
-        // Type::NULLVALUE is always false
+        case Value::Type::__EMPTY__:
         case Value::Type::NULLVALUE: {
-            return std::make_pair(false, true);
+            return Value::kNullValue;
         }
         case Value::Type::BOOL: {
-            return std::make_pair(getBool(), true);
-        }
-        case Value::Type::INT: {
-            return std::make_pair(getInt() != 0, true);
-        }
-        case Value::Type::FLOAT: {
-            return std::make_pair(std::abs(getFloat()) > kEpsilon, true);
+            return *this;
         }
         case Value::Type::STRING: {
-            return std::make_pair(!getStr().empty(), true);
-        }
-        case Value::Type::DATE: {
-            return std::make_pair(getDate().toInt() != 0, true);
+            auto str = toString();
+            std::transform(str.begin(), str.end(), str.begin(), ::tolower);
+            if (str.compare("true") == 0) {
+                return true;
+            }
+            if (str.compare("false") == 0) {
+                return false;
+            }
+            return Value::kNullValue;
         }
         default: {
-            return std::make_pair(bool{}, false);
+            return Value::kNullBadType;
         }
     }
 }
 
-std::pair<double, bool> Value::toFloat() {
+Value Value::toFloat() {
     switch (type_) {
         case Value::Type::INT: {
-            return std::make_pair(static_cast<double>(getInt()), true);
+            return static_cast<double>(getInt());
         }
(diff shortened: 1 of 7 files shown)


src/common/datatypes/Value.cpp
```diff
@@ -1550,79 +1550,72 @@ std::string Value::toString() const {
...
         }
         case Value::Type::STRING: {
-            return std::make_pair(!getStr().empty(), true);
-        }
-        case Value::Type::DATE: {
-            return std::make_pair(getDate().toInt() != 0, true);
+            if (toString().compare("true") == 0) {
+                return true;
+            }
+            if (toString().compare("false") == 0) {
```
yixinglu, line 1566: you'd better to assign the result of `toString()` to a variable for avoiding this function to be called multiple times.

czpmango: Fixed, thx.

src/common/datatypes/Value.cpp
```diff
@@ -1550,79 +1550,74 @@ std::string Value::toString() const {
...
-            return std::make_pair(std::abs(getFloat()) > kEpsilon, true);
+            return *this;
         }
         case Value::Type::STRING: {
-            return std::make_pair(!getStr().empty(), true);
-        }
-        case Value::Type::DATE: {
-            return std::make_pair(getDate().toInt() != 0, true);
+            auto str = toString();
+            std::transform(str.begin(), str.end(), str.begin(), ::tolower);
```
jievince, line 1564: why not directly call std::lower(toString())?

czpmango: std::tolower?

jievince: I'am sorry. Seems std::string doesn't has tolower()

src/common/datatypes/Value.cpp
```diff
@@ -1550,79 +1550,74 @@ std::string Value::toString() const {
     LOG(FATAL) << "Unknown value type " << static_cast<int>(type_);
 }
 
-std::pair<bool, bool> Value::toBool() {
+Value Value::toBool() {
```
panda-sheep, line 1553: please convert to bool in this function

panda-sheep (changes requested):
please convert Value::toBool  to bool

src/common/datatypes/Value.cpp
czpmango: You could get the bool type return value explicitly in the caller interface.

panda-sheep: I don’t think so, I think you should check the original function implementation

critical27:
> - 字符串 "true"/"FaLse" 等会返回转换后的 bool 值

Why do you prefer literally true/false to empty?

czpmango:
> > * 字符串 "true"/"FaLse" 等会返回转换后的 bool 值
> 
> Why do you prefer literally true/false to empty?

Consistent with the openCypher standard.

neo4j test:
```
neo4j@neo4j> return toBoolean("FaLse");
+--------------------+
| toBoolean("FaLse") |
+--------------------+
| FALSE              |
+--------------------+
```

panda-sheep (dismissed):
Good job，lgtm

panda-sheep (approved):

panda-sheep (approved):

yixinglu (approved):

PR #414, opened 2021-02-18
reveal the hidden leader changed error code

liuyu85cn:
when leader changed, 

collectResponse only set error code to StorageRpcResponse::responses_

but not append error code to failedParts_.

Which means if the caller not check responses_, it may get a succeeded result.


panda-sheep (approved):
Good job！

critical27 (approved):

PR #415, opened 2021-02-19
Fix the value of sha in http server.

Shylock-Hg:


laura-ding (commented):
The graph and the storage will use the same git info

Shylock-Hg:
> The graph and the storage will use the same git info

Won't

yixinglu (approved):

czpmango (approved):

PR #418, opened 2021-02-23
Fix nonegative comparison.

Shylock-Hg:


jievince (dismissed):

src/common/function/FunctionManager.cpp
```diff
@@ -890,9 +890,7 @@ FunctionManager::FunctionManager() {
             if (args[0].isStr() && args[1].isInt() && args[2].isStr()) {
                 auto value = args[0].getStr();
                 size_t size = args[1].getInt();
```
yixinglu, line 892: It's better to change the type of size to `int64_t` rather than `size_t`

src/common/function/FunctionManager.cpp
```diff
@@ -1558,8 +1556,7 @@ FunctionManager::FunctionManager() {
                     std::distance(colNames.begin(),
                                   std::find(colNames.begin(), colNames.end(), args[2].getStr()));
             }
-            if ((rowIndex >= ds.rowSize() || rowIndex < 0) ||
-                (colIndex >= ds.colSize() || colIndex < 0)) {
+            if (rowIndex >= ds.rowSize() || colIndex >= ds.colSize()) {
```
yixinglu, line 1559: ditto for `rowIndex` and `colIndex`

src/common/function/FunctionManager.cpp
Shylock-Hg: Fixed.

nevermore3 (approved):

laura-ding (approved):

PR #419, opened 2021-02-24
Add expired time factor.

Shylock-Hg:


critical27:
```
    def _format_nebula_command(self, name, meta_port, ports, debug_log=True):
        param_format = "--meta_server_addrs={} --port={} --ws_http_port={} --ws_h2_port={} --heartbeat_interval_secs=1"
```

Is this code still working? If it is, it seems we don't need to change it. Just unify all heartbeat_interval_secs in ci conf.

Shylock-Hg:
> ```
>     def _format_nebula_command(self, name, meta_port, ports, debug_log=True):
>         param_format = "--meta_server_addrs={} --port={} --ws_http_port={} --ws_h2_port={} --heartbeat_interval_secs=1"
> ```
> 
> Is this code still working? If it is, it seems we don't need to change it. Just unify all heartbeat_interval_secs in ci conf.

We need to expire based on heartbeat time.

Shylock-Hg:
> what is `expired_time_factor` mean? will using `heartbeat_interval_secs` * `expired_time_factor` to determine timeout ?

Yes.

critical27:
> heartbeat_interval_secs

The ci fail, probably because meta/storage `heartbeat_interval_secs` is not unified.

Shylock-Hg:
> > --heartbeat_interval_secs=1
> 
> The ci fail, probably because meta/storage `heartbeat_interval_secs` is not unified.
> `heartbeat_interval_secs * 2`has been used quite a looooog time, since 1.0.

1. The heartbeat config in ic is same for meta/storage.
2. I see the origin expired value in meta 
```
DEFINE_int32(expired_threshold_sec, 10 * 60,
             "Hosts will be expired in this time if no heartbeat received");
```
I think it's 600s now it's heartbeat*2 (2s in ci in fact).

darionyaphet:
> > what is `expired_time_factor` mean? will using `heartbeat_interval_secs` * `expired_time_factor` to determine timeout ?
> 
> Yes.

What is the benefit for this ？

Shylock-Hg:
> > > what is `expired_time_factor` mean? will using `heartbeat_interval_secs` * `expired_time_factor` to determine timeout ?
> > 
> > 
> > Yes.
> 
> What is the benefit for this ？

Could fit the environment (e.g. If running on slow environment the user could set a big expired_time_factor to avoid mistake)

critical27:
> expired_threshold_sec

The root cause is `heartbeat_interval_secs` in meta is 1s, and storage's `heartbeat_interval_secs` is not 1s. And that will make hosts lost.

laura-ding (approved):

nevermore3 (approved):

PR #422, opened 2021-02-26
Add bound check to int() and update tests for type casting

Aiee:
As title.

### DIFF
--- a/src/common/datatypes/Value.cpp
+++ b/src/common/datatypes/Value.cpp
@@ -1610,14 +1610,10 @@ std::pair<int64_t, bool> Value::toInt() {
             return std::make_pair(getInt(), true);
         }
         case Value::Type::FLOAT: {
-            // Check if float value is in the limit of int_64
-            if (getFloat() < std::numeric_limits<int64_t>::min()) {
-                LOG(WARNING) << "Overflow happened when casting float value " << getFloat()
-                             << " into int";
+            // Check if float value is in the range of int_64
+            if (getFloat() <= std::numeric_limits<int64_t>::min()) {
                 return std::make_pair(std::numeric_limits<int64_t>::min(), false);
-            } else if (getFloat() > std::numeric_limits<int64_t>::max()) {
-                LOG(WARNING) << "Overflow happened when casting float value " << getFloat()
-                             << " into int";
+            } else if (getFloat() >= std::numeric_limits<int64_t>::max()) {
                 return std::make_pair(std::numeric_limits<int64_t>::max(), false);
             }
             return std::make_pair(static_cast<int64_t>(getFloat()), true);
--- a/src/common/datatypes/Value.cpp
+++ b/src/common/datatypes/Value.cpp
@@ -1610,6 +1610,16 @@ std::pair<int64_t, bool> Value::toInt() {
             return std::make_pair(getInt(), true);
         }
         case Value::Type::FLOAT: {
+            // Check if float value is in the limit of int_64
+            if (getFloat() < std::numeric_limits<int64_t>::min()) {
+                LOG(WARNING) << "Overflow happened when casting float value " << getFloat()
+                             << " into int";
+                return std::make_pair(std::numeric_limits<int64_t>::min(), false);
+            } else if (getFloat() > std::numeric_limits<int64_t>::max()) {
+                LOG(WARNING) << "Overflow happened when casting float value " << getFloat()
+                             << " into int";
+                return std::make_pair(std::numeric_limits<int64_t>::max(), false);
+            }
             return std::make_pair(static_cast<int64_t>(getFloat()), true);
         }
         case Value::Type::STRING: {
--- a/src/common/datatypes/Value.cpp
+++ b/src/common/datatypes/Value.cpp
@@ -1611,10 +1611,11 @@ std::pair<int64_t, bool> Value::toInt() {
         }
         case Value::Type::FLOAT: {
             // Check if float value is in the range of int_64
+            // Return min/max int_64 value and false to accommodate Cypher
             if (getFloat() <= std::numeric_limits<int64_t>::min()) {
-                return std::make_pair(std::numeric_limits<int64_t>::min(), false);
+                return std::make_pair(std::numeric_limits<int64_t>::min(), true);
             } else if (getFloat() >= std::numeric_limits<int64_t>::max()) {
-                return std::make_pair(std::numeric_limits<int64_t>::max(), false);
+                return std::make_pair(std::numeric_limits<int64_t>::max(), true);
             }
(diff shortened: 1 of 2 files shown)


jievince (dismissed):

src/common/datatypes/Value.cpp
```diff
@@ -1610,6 +1610,11 @@ std::pair<int64_t, bool> Value::toInt() {
             return std::make_pair(getInt(), true);
         }
         case Value::Type::FLOAT: {
+            // Check if float value is in the limit of int_64
+            if (getFloat() < std::numeric_limits<int64_t>::min() ||
+                getFloat() > std::numeric_limits<int64_t>::max()) {
+                    return std::make_pair(int64_t{}, false);
```
Shylock-Hg, line 1616: The cypher will return min/max value of int64_t when overflow.

czpmango, line 1616: Why add this logic? The numerical value will be truncated when converted to int.

Aiee: Thanks for pointing that. I have updated the value returned in a overflow.

Aiee: I believe the user should be somehow informed when an overflow happens. Returning a truncated value without notifying the user may cause confusion.

Shylock-Hg: static_cast is ok for this logical. You could remain the tests.

src/common/datatypes/Value.cpp
```diff
@@ -1610,6 +1610,16 @@ std::pair<int64_t, bool> Value::toInt() {
             return std::make_pair(getInt(), true);
         }
         case Value::Type::FLOAT: {
+            // Check if float value is in the limit of int_64
+            if (getFloat() < std::numeric_limits<int64_t>::min()) {
+                LOG(WARNING) << "Overflow happened when casting float value " << getFloat()
```
Shylock-Hg, line 1615: And don't log in so low-level layer it maybe hurt performance.

src/common/datatypes/Value.cpp
czpmango: I think your consideration is reasonable.
This is my test in mysql:
```
mysql> select CONVERT((-9223372036854775809.3),unsigned);
+--------------------------------------------+
| CONVERT((-9223372036854775809.3),unsigned) |
+--------------------------------------------+
|                        9223372036854775808 |
+--------------------------------------------+
1 row in set, 1 warning (0.00 sec)

mysql> show warnings;
+---------+------+-------------------------------------------------------------+
| Level   | Code | Message                                                     |
+---------+------+-------------------------------------------------------------+
| Warning | 1292 | Truncated incorrect DECIMAL value: '-9223372036854775809.3' |
+---------+------+-------------------------------------------------------------+
1 row in set (0.00 sec)
```

src/common/datatypes/Value.cpp
czpmango: > And don't log in so low-level layer it maybe hurt performance.

Agree.

src/common/datatypes/Value.cpp
Aiee: So theoretically converting a float into an int is always valid, is it correct?

src/common/datatypes/Value.cpp
Aiee: Makes sense. I'll log the warning when calling this function.

src/common/datatypes/Value.cpp
Aiee: Good to know about this. Thanks.

Aiee:
How about this approach:
In an overflow, toInt() returns the ```truncated result``` as the first element of the pair, and ```false``` as the second so that we can log warnings base on this boolean value.

WDYT? @Shylock-Hg @czpmango

src/common/datatypes/Value.cpp
Shylock-Hg: Could keep the check.It's maybe a U.B.

Shylock-Hg:
> How about this approach:
> In an overflow, toInt() returns the `truncated result` as the first element of the pair, and `false` as the second so that we can log warnings base on this boolean value.
> 
> WDYT? @Shylock-Hg @czpmango

Make sense.

src/common/datatypes/Value.cpp
```diff
@@ -1610,6 +1610,12 @@ std::pair<int64_t, bool> Value::toInt() {
             return std::make_pair(getInt(), true);
         }
         case Value::Type::FLOAT: {
+            // Check if float value is in the limit of int_64
+            if (getFloat() < std::numeric_limits<int64_t>::min() ||
```
Shylock-Hg, line 1614: Less and equal

src/common/datatypes/Value.cpp
```diff
@@ -1610,6 +1610,12 @@ std::pair<int64_t, bool> Value::toInt() {
...
+                getFloat() > std::numeric_limits<int64_t>::max()) {
```
Shylock-Hg, line 1615: ditto.

src/common/datatypes/Value.cpp
```diff
@@ -1610,6 +1610,12 @@ std::pair<int64_t, bool> Value::toInt() {
...
+                // return a truncated value when overflow
+                return std::make_pair(static_cast<int64_t>(getFloat()), false);
```
Shylock-Hg, line 1617: Don't cast it, it's maybe a U.B.

Aiee: Fixed

src/common/datatypes/Value.cpp
```diff
@@ -1610,6 +1610,12 @@ std::pair<int64_t, bool> Value::toInt() {
             return std::make_pair(getInt(), true);
         }
         case Value::Type::FLOAT: {
+            // Check if float value is in the range of int_64
+            if (getFloat() <= std::numeric_limits<int64_t>::min()) {
+                return std::make_pair(std::numeric_limits<int64_t>::min(), false);
```
Shylock-Hg, line 1615: return false means the conversion is invalid, but Cypher allow this conversion even if overflow.

Aiee: if we return true for overflow then how could we distinguish an overflow and a non-overflow situation?

Shylock-Hg: Will cypher do it?

Aiee: Screen Shot 2021-03-02 at 11 43 25 AM

cypher returns the int_min/max without logging

Shylock-Hg: Following the cypher.

Shylock-Hg (approved):

czpmango (approved):

yixinglu (approved):

PR #423, opened 2021-03-01
increate the range of storage error code

liuyu85cn:
the previous 100 error code is hard to group by different kinds


liuyu85cn:
> the previous 100 error code is hard to group by different kinds   

actually, I thought, may be the largest "E_ERR_UNKOWN" should be 999999

and graph use 1xxxxx

meta use 2xxxxx

storage use 3xxxxx

PR #424, opened 2021-03-01
always send storage leaders to meta

critical27:
1. Previously, leader info only send to meta when it changed, which will cause `show host` or `show parts` is missing something if heartbeat with leader info failed. Talked with @sherman-the-tank offline, always send is ok. 
2. Unify default value of `heartbeat_interval_secs` in gflags and conf.

### DIFF
--- a/src/common/clients/meta/MetaClient.cpp
+++ b/src/common/clients/meta/MetaClient.cpp
@@ -19,7 +19,7 @@
 #include <folly/executors/Async.h>
 #include <folly/futures/Future.h>
 
-DEFINE_int32(heartbeat_interval_secs, 3, "Heartbeat interval");
+DEFINE_int32(heartbeat_interval_secs, 10, "Heartbeat interval");
 DEFINE_int32(meta_client_retry_times, 3, "meta client retry times, 0 means no retry");
 DEFINE_int32(meta_client_retry_interval_secs, 1,
              "meta client sleep interval between retry");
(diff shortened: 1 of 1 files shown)


src/common/clients/meta/MetaClient.cpp
```diff
@@ -2297,8 +2297,8 @@ folly::Future<StatusOr<bool>> MetaClient::heartbeat() {
                     leaderIds_.clear();
                     leaderIds_ = leaderIds;
                 }
-                req.set_leader_partIds(std::move(leaderIds));
             }
+            req.set_leader_partIds(std::move(leaderIds));
```
bright-starry-sky, line 2301: What are the advantages?

bright-starry-sky, line 2301: What’s the advantage?

panda-sheep: Good
In the past, if `leaderIds_ = leaderIds` was executed and the subsequent `future_heartBeat `failed to execute, would the execution fail forever before the next leader change?

critical27: In worst case, if this heartbeat carry leader info, and RPC failed because of network or something else, meta will never get this storaged's leader info

bright-starry-sky (approved):
LGTM

panda-sheep (approved):
Good job!

PR #428, opened 2021-03-05
Invalid host for host choosen.

Shylock-Hg:
NG-460
In the case if one host offline, the first request to it will failed and remove it from leaders, the later requests will choose random host from all host contains corresponding partition even the offline host, so it maybe failed too even if the new leader had elected.
Re-product by:
1. Setup up cluster,
2. insert data (sync leaders)
3. offline one host
4. insert data to offline partition
5. try 4 again

The invalid hosts maybe can't valid again so remained the random choosing logical still.

### DIFF
--- a/src/common/clients/storage/StorageClientBase.inl
+++ b/src/common/clients/storage/StorageClientBase.inl
@@ -123,7 +123,21 @@ StorageClientBase<ClientType>::getLeader(const meta::PartHosts& partHosts) const
     }
     {
         folly::RWSpinLock::WriteHolder wh(leadersLock_);
-        VLOG(1) << "No leader exists. Choose one random.";
+        VLOG(1) << "No leader exists. Choose one from valid hosts.";
+        {
+            folly::RWSpinLock::ReadHolder rh(hostsLock_);
+            for (const auto &host : partHosts.hosts_) {
+                if (invalidHosts_.find(host) != invalidHosts_.end()) {
+                    leaders_[part] = host;
+                    break;
+                }
+            }
+        }
+        auto find = leaders_.find(part);
+        if (find != leaders_.end()) {
+            return find->second;
+        }
+        VLOG(1) << "Maybe no valid leader exists. Choose random one for a try.";
         const auto& random = partHosts.hosts_[folly::Random::rand32(partHosts.hosts_.size())];
         leaders_[part] = random;
         return random;
@@ -165,6 +179,17 @@ void StorageClientBase<ClientType>::invalidLeader(GraphSpaceID spaceId,
     }
 }
 
+template<typename ClientType>
+void StorageClientBase<ClientType>::invalidHost(const HostAddr &host) {
+    folly::RWSpinLock::WriteHolder wh(hostsLock_);
+    invalidHosts_.emplace(host);
+}
+
+template<typename ClientType>
+void StorageClientBase<ClientType>::validHost(const HostAddr &host) {
+    folly::RWSpinLock::WriteHolder wh(hostsLock_);
+    invalidHosts_.erase(host);
+}
 
 template<typename ClientType>
 template<class Request, class RemoteFunc, class Response>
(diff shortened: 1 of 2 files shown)


src/common/clients/storage/StorageClientBase.inl
```diff
@@ -123,7 +123,21 @@ StorageClientBase<ClientType>::getLeader(const meta::PartHosts& partHosts) const
     }
     {
         folly::RWSpinLock::WriteHolder wh(leadersLock_);
-        VLOG(1) << "No leader exists. Choose one random.";
+        VLOG(1) << "No leader exists. Choose one from valid hosts.";
+        {
+            folly::RWSpinLock::ReadHolder rh(hostsLock_);
+            for (const auto &host : partHosts.hosts_) {
+                if (invalidHosts_.find(host) != invalidHosts_.end()) {
```
panda-sheep, line 130: `if (invalidHosts_.find(host) == invalidHosts_.end())`  ？

src/common/clients/storage/StorageClientBase.inl
```diff
@@ -212,9 +237,11 @@ StorageClientBase<ClientType>::collectResponse(
                                << " failed: " << val.exception().what();
                     auto parts = getReqPartsId(r);
                     context->resp.appendFailedParts(parts, storage::cpp2::ErrorCode::E_RPC_FAILURE);
+                    invalidHost(host);
```
panda-sheep, line 240: If rpc fails once, is this host invalid?

panda-sheep: If there are 3 copies, if each machine fails the rpc once due to network problems,
It will never be able to elect a leader?

src/common/clients/storage/StorageClientBase.inl
Shylock-Hg: Not found mean valid.

src/common/clients/storage/StorageClientBase.inl
Shylock-Hg: For once failed, the second request will send to replica(get correct leader) and then re-send to leader.

src/common/clients/storage/StorageClientBase.inl
panda-sheep: Isn't this equal to end（）? This is found

src/common/clients/storage/StorageClientBase.inl
panda-sheep: The first rpc fails, host is added to invalidHost.
Rpc failed for the second time, and host was added to invalidHost.
Rpc failed for the third time, and host was added to invalidHost.
What about the fourth retry? Can be used without a host

panda-sheep (commented):
The most fundamental point: rpc fails, host is not equal to invalid

critical27:
Have you test start graph at first, and start 3 storage later? Can this work?

src/common/clients/storage/StorageClientBase.inl
Shylock-Hg: The second request will get leader and the third request will send to the leader.

critical27:
> Why not use round robin mode for host selection?

Exactly, we have talked about this offline before

Shylock-Hg:
> > Why not use round robin mode for host selection?
> 
> Exactly, we have talked about this offline before

Yes, but I think round robin mode can't avoid the invalid host too.

critical27:
> > > Why not use round robin mode for host selection?
> > 
> > 
> > Exactly, we have talked about this offline before
> 
> Yes, but I think round robin mode can't avoid the invalid host too.

The invalid host is added by rpc failure, this PR only can handle killed storage. But the problem is that when all services are healthy, just network interruption, it is more general to round robin.

Shylock-Hg:
> Have you test start graph at first, and start 3 storage later? Can this work?

I tried this case but noting unexpected, after leader elected the request works well.

critical27:
> > Have you test start graph at first, and start 3 storage later? Can this work?
> 
> I tried this case but nothing unexpected, after leader elected the request works well.

...

src/common/clients/storage/StorageClientBase.inl
critical27: > Not found mean valid.

Found mean invalid, so why you set `leaders_[part] = host;`

Shylock-Hg: Good catch. Fixed.

PR #438, opened 2021-03-10
select leader in round-robin

critical27:
1. When we don't know who is leader, select in round-robin. Previously random will make us need to retry several times(some query like `lookup` need to access all leaders), until we random pick a live storage for all partition (which is hard if we have a lot partitions, I tried for 10+ times when 100 partitions). In round-robin ways, we only need to retry for a limited times if we encounter leader change.

### DIFF
--- a/src/common/clients/meta/MetaClient.cpp
+++ b/src/common/clients/meta/MetaClient.cpp
@@ -2991,7 +2991,7 @@ Status MetaClient::refreshCache() {
 }
 
 
-StatusOr<LeaderMap> MetaClient::loadLeader() {
+StatusOr<LeaderInfo> MetaClient::loadLeader() {
     // Return error if has not loadData before
     if (!ready_) {
         return Status::Error("Not ready!");
@@ -3002,8 +3002,9 @@ StatusOr<LeaderMap> MetaClient::loadLeader() {
         return Status::Error("List hosts failed");
     }
 
-    LeaderMap leaderMap;
+    LeaderInfo leaderInfo;
     auto hostItems = std::move(ret).value();
+    std::unordered_map<GraphSpaceID, size_t> leaderCount;
     for (auto& item : hostItems) {
         for (auto& spaceEntry : item.get_leader_parts()) {
             auto spaceName = spaceEntry.first;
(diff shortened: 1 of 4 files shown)


src/common/clients/storage/StorageClientBase.h
```diff
@@ -248,7 +248,9 @@ class StorageClientBase {
 
     mutable folly::RWSpinLock leadersLock_;
     mutable std::unordered_map<std::pair<GraphSpaceID, PartitionID>, HostAddr> leaders_;
-    mutable std::atomic_bool loadLeaderBefore_{false};
+    // record the index of hosts we pick last time
+    mutable std::unordered_map<std::pair<GraphSpaceID, PartitionID>, size_t> leaderIndex_;
```
Shylock-Hg, line 252: I think you should initialize the `leaderIndex_` when load leader first.

src/common/clients/meta/MetaClient.cpp
```diff
@@ -3013,14 +3014,27 @@ StatusOr<LeaderMap> MetaClient::loadLeader() {
...
+                    for (size_t i = 0; i < peers.size(); i++) {
+                        if (peers[i] == item.hostAddr) {
+                            leaderIndex = i;
+                            break;
+                        }
+                    }
+                }
+                leaderInfo.leaderIndex_[{spaceId, partId}] = leaderIndex;
             }
+            leaderCount[spaceId] += spaceEntry.second.size();
```
bright-starry-sky, line 3031: What's the purpose of `leaderCount` ?

critical27: I forgot to delete it, I check it before, I think it will be dangerous. Let me delete it.

src/common/clients/storage/StorageClientBase.h
critical27: Make sense, fixed

bright-starry-sky (approved):
Nice job.

CPWstatic (approved):

bright-starry-sky (approved):

PR #443, opened 2021-03-12
add exist expression

nevermore3:


src/common/datatypes/Edge.cpp
```diff
@@ -32,6 +32,58 @@ std::string Edge::toString() const {
...
+    }
+    if (dst != rhs.dst) {
+        return dst < rhs.dst;
+    }
+    if (type != rhs.type) {
+        return type < rhs.type;
+    }
+    if (ranking != rhs.ranking) {
+        return ranking < rhs.ranking;
+    }
```
yixinglu, line 54: Why not unify the direction of edge and then compare their src/dst/ranking ?

src/common/datatypes/Vertex.cpp
```diff
@@ -25,6 +25,57 @@ std::string Tag::toString() const {
...
+}
+
+bool Vertex::operator<(const Vertex& rhs) const {
+    if (vid != rhs.vid) {
+        return vid < rhs.vid;
+    }
+    if (tags.size() != rhs.tags.size()) {
+        return tags.size() < rhs.tags.size();
+    }
+    return false;
```
yixinglu, line 76: Why not compare the properties of vertex tags?

yixinglu (approved):

jievince (approved):

PR #455, opened 2021-03-17
Feature/fetch storage leader by heartbeat

Shylock-Hg:
Now the server is ready when
1. first elected and one heartbeat more about time
2. leader changed : elected and offline host expired and one heartbeat more about time

Require https://github.com/vesoft-inc/nebula-storage/pull/393

### DIFF
--- a/src/common/clients/meta/MetaClient.cpp
+++ b/src/common/clients/meta/MetaClient.cpp
@@ -252,9 +252,6 @@ bool MetaClient::loadData() {
     }
 
     auto leaderMap = loadLeader(spaceIndexByName, cache);
-    if (!leaderMap.ok()) {
-        return false;
-    }
 
     decltype(localCache_) oldCache;
     {
@@ -269,13 +266,19 @@ bool MetaClient::loadData() {
         spaceEdgeIndexByType_   = std::move(spaceEdgeIndexByType);
         spaceTagIndexById_      = std::move(spaceTagIndexById);
         spaceAllEdgeMap_        = std::move(spaceAllEdgeMap);
-        leaderMap_              = std::move(leaderMap).value();
+        if (leaderMap.ok()) {
+            leaderMap_              = leaderMap.value();
+        }
     }
 
     diff(oldCache, localCache_);
     listenerDiff(oldCache, localCache_);
     loadRemoteListeners();
     ready_ = true;
+    // TODO(shylock) using separate timestamp to avoid conflict with meta data
+    if (!leaderMap.ok()) {
+        return false;
+    }
     return true;
 }
 
--- a/src/common/clients/meta/MetaClient.cpp
+++ b/src/common/clients/meta/MetaClient.cpp
@@ -251,6 +251,11 @@ bool MetaClient::loadData() {
         spaceIndexByName.emplace(space.second, spaceId);
     }
 
+    auto leaderMap = loadLeader(spaceIndexByName, cache);
+    if (!leaderMap.ok()) {
+        return false;
+    }
+
     decltype(localCache_) oldCache;
     {
         folly::RWSpinLock::WriteHolder holder(localCacheLock_);
@@ -264,6 +269,7 @@ bool MetaClient::loadData() {
         spaceEdgeIndexByType_   = std::move(spaceEdgeIndexByType);
         spaceTagIndexById_      = std::move(spaceTagIndexById);
         spaceAllEdgeMap_        = std::move(spaceAllEdgeMap);
+        leaderMap_              = std::move(leaderMap).value();
     }
 
     diff(oldCache, localCache_);
@@ -543,7 +549,7 @@ void MetaClient::getResponse(Request req,
                      this] () mutable {
         auto client = clientsMan_->client(host, evb, false, FLAGS_meta_client_timeout_ms);
         VLOG(1) << "Send request to meta " << host;
-        remoteFunc(client, req)
+        remoteFunc(DCHECK_NOTNULL(client), req)
             .via(evb)
             .then([host,
                    req = std::move(req),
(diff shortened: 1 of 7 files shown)


src/common/clients/meta/MetaClient.cpp
```diff
@@ -2991,34 +3023,45 @@ Status MetaClient::refreshCache() {
...
+            auto spaceId = find->second;
+            leaderCount[spaceId] += spaceEntry.second.size();
             for (const auto& partId : spaceEntry.second) {
                 leaderMap[{spaceId, partId}] = item.hostAddr;
             }
         }
         LOG(INFO) << "Load leader of " << item.hostAddr
                   << " in " << item.get_leader_parts().size() << " space";
     }
+    // check leader count
```
critical27, line 3052: This check here is a little aggressive here? For example, if a storage is dead (one replica), all request will be failed (even the request not query the dead one) because the `leaderMap` is not complete. I think we could just use the leaderMap, and retry later.

src/common/clients/storage/StorageClientBase.inl
```diff
@@ -217,19 +152,6 @@ StorageClientBase<ClientType>::collectResponse(
                                 << ", failed code " << static_cast<int32_t>(code.get_code());
                         hasFailure = true;
                         context->resp.emplaceFailedPart(code.get_part_id(), code.get_code());
-                        if (code.get_code() == storage::cpp2::ErrorCode::E_LEADER_CHANGED) {
-                            auto* leader = code.get_leader();
-                            if (isValidHostPtr(leader)) {
-                                updateLeader(spaceId, code.get_part_id(), *leader);
```
critical27, line 223: I think invalidLeader could be removed. But `updateLeader` could be reserved, because storage has explicitly tell who is leader. Even fetch info from meta, it is the same result. WDYT?

critical27 (commented):
Generally LGTM, there are some questions, please see the comments

src/common/clients/meta/MetaClient.cpp
Shylock-Hg: If the leader is not completed, the leader cache in metaclient won't update, it keep same as before.

src/common/clients/storage/StorageClientBase.inl
Shylock-Hg: Yes, but I think it's meaningless, which reduce some partition wait time. And if we enable both retry and heartbeat mechanism, it will be more complex and will introduce some corner case.

src/common/clients/meta/MetaClient.cpp
critical27: Then how to query when leader cache is empty and leader count is not complete?

Shylock-Hg: The request will failed for can't found leader.

critical27: This is unacceptable, only a part of cluster is dead should not have impacts on the whole cluster.

Shylock-Hg: If so, we could accept it when leader count <= partition count.

critical27:
@yixinglu @CPWstatic @bright-starry-sky PTAL

src/common/clients/meta/MetaClient.cpp
bright-starry-sky: emm. I think sometimes invalid leaders are allowed, the purpose of this PR is to minimize the failure rate, rather than solve the problem completely. WDYT?

Shylock-Hg: Do you means don't check the leader count ?

bright-starry-sky: Yes, that's what I mean. WDYT?

Shylock-Hg: It's not workable which I discussed with @critical27 that if don't check the leader count the meta client never know some host expired so the meta client will keep the expired host always.

bright-starry-sky:
Generally LGTM

src/common/clients/storage/StorageClientBase.inl
critical27: There won't be much corner cases. Just copy a leaderMap from meta, and modify it with protection of a mutex like before.

src/common/clients/meta/MetaClient.cpp
critical27: > If so, we could accept it when leader count <= partition count.

Yeah, I think `leader count <= partition count` is reasonable.

critical27: > Yes, that's what I mean. WDYT?

There is a scenario that, say we have 100 parts of 3 replicas. Leader count is [30, 30, 40],  we kill the last one, after leader change, total leader count from meta may be greater than 100 (the dead one is still regard as online for a short period of time), it is possible that we see [50, 50, 40] from meta. 

After some time, leader count from meta will be [50, 50, 0], but the update time won't be update since we don't check it in meta.

src/common/clients/storage/StorageClientBase.inl
panda-sheep: So I think the leader contained in the response returned by storage is meaningless?

Shylock-Hg: I think it's meaning less if we con't use the retry mechanism.

src/common/clients/meta/MetaClient.cpp
bright-starry-sky: make sense. sometimes, the total of leaders is less than the actual number of partitions is more meaningful.

src/common/clients/storage/StorageClientBase.inl
critical27: In worst case we need extra `2 * heartbeat` to get leader map, why this is meaningless? By average, it is still 10 seconds by default.

panda-sheep: In this pr, the leader information is obtained directly from metaclient,
The leader information returned by the storage response is useless?

Shylock-Hg:
@critical27 take a view the new commits.

critical27 (approved):
Good job, LGTM

critical27:
@Shylock-Hg some storage code use the `MetaClient::loadLeader`, would you apply the change to https://github.com/vesoft-inc/nebula-storage/pull/393 as well later.

Shylock-Hg:
> @Shylock-Hg some storage code use the `MetaClient::loadLeader`, would you apply the change to [vesoft-inc/nebula-storage#393](https://github.com/vesoft-inc/nebula-storage/pull/393) as well later.

Fixed.

PR #459, opened 2021-03-18
Support get neighbors with steps

darionyaphet:


src/common/clients/storage/StorageClientBase.inl
```diff
@@ -376,6 +376,38 @@ StorageClientBase<ClientType>::clusterIdsToHosts(GraphSpaceID spaceId,
     return clusters;
 }
 
+template<typename ClientType>
+template<class GetIdFunc>
+StatusOr<std::unordered_map<HostAddr,
+                            std::unordered_map<
+                                PartitionID,
+                                storage::cpp2::StepRows>
+                            >>
```
yixinglu, line 385: why not to define this type's alias for easy usage.

yixinglu (commented):
impressive interface changes for graph usage. Good job.

PR #461, opened 2021-03-19
remove reduncant parentheses

nevermore3:


laura-ding (approved):

laura-ding:
The cpplint check failed, please solve it.

Aiee (approved):

Shylock-Hg:
If we want to provide the expression literal same with user input query better to store the origin expression literal in parser like mysql.

PR #467, opened 2021-03-23
Add agg cache

czpmango:
> 

Add aggregateExpression's runtime cache.Just Follow [https://github.com/vesoft-inc/nebula-common/pull/407](https://github.com/vesoft-inc/nebula-common/pull/407) to achieve this.

---

Benchmark:
Before this PR:
c1

After this PR:
c2

### DIFF
--- a/src/common/expression/AggregateExpression.cpp
+++ b/src/common/expression/AggregateExpression.cpp
@@ -30,6 +30,10 @@ void AggregateExpression::resetFrom(Decoder& decoder) {
     name_ = decoder.readStr();
     distinct_ = decoder.readValue().getBool();
     arg_ = decoder.readExpression();
+    auto aggFuncResult = AggFunctionManager::get(*DCHECK_NOTNULL(name_));
+    if (aggFuncResult.ok()) {
+        aggFunc_ = std::move(aggFuncResult).value();
+    }
 }
 
 const Value& AggregateExpression::eval(ExpressionContext& ctx) {
@@ -43,20 +47,17 @@ const Value& AggregateExpression::eval(ExpressionContext& ctx) {
         uniques->values.emplace(val);
     }
 
-    apply(aggData_, val);
+    DCHECK(aggFunc_);
+    aggFunc_(aggData_, val);
 
     return aggData_->result();
-}
-
-void AggregateExpression::apply(AggData* aggData, const Value& val) {
-    AggFunctionManager::get(*name_).value()(aggData, val);
 }
 
 std::string AggregateExpression::toString() const {
     // TODO fix it
     std::string arg;
     if (arg_->kind() == Expression::Kind::kConstant) {
-        const auto *argExpr = static_cast<const ConstantExpression*>(arg_.get());
+        const auto* argExpr = static_cast<const ConstantExpression*>(arg_.get());
         if (argExpr->value().isStr()) {
             arg = argExpr->value().getStr();
         } else {
@@ -77,4 +78,4 @@ std::string AggregateExpression::toString() const {
 void AggregateExpression::accept(ExprVisitor* visitor) {
     visitor->visit(this);
 }
-}  // namespace nebula
+}   // namespace nebula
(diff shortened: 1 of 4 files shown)


src/common/expression/AggregateExpression.cpp
```diff
@@ -43,15 +47,12 @@ const Value& AggregateExpression::eval(ExpressionContext& ctx) {
         uniques->values.emplace(val);
     }
 
-    apply(aggData_, val);
+    DCHECK(aggFunc_.hasValue());
```
Shylock-Hg, line 50: How about invalid agg function name?

czpmango: Checked in parser or validator. Invalid function name is not expected here.

Shylock-Hg: Do we have the case in tck?

czpmango: Do you mean this case:
```
(czp@nebula) [nba]> go from "Tim Duncan" over like yield $$.player.age as age | group by $-.age yield $-.age,count($-.age) as count
+--------+-------+
| $-.age | count |
+--------+-------+
| 38     | 2     |
+--------+-------+
| 41     | 1     |
+--------+-------+
| 36     | 1     |
+--------+-------+
Got 3 rows (time spent 387690/395754 us)

Wed, 24 Mar 2021 13:59:12 CST

(czp@nebula) [nba]> go from "Tim Duncan" over like yield $$.player.age as age | group by $-.age yield $-.age,counte($-.age) as count
[ERROR (-7)]: SyntaxError: Unknown function  near `counte'
```

Shylock-Hg: Yes, it's ok for me

Shylock-Hg (approved):

src/common/expression/AggregateExpression.h
```diff
@@ -78,15 +84,16 @@ class AggregateExpression final : public Expression {
...
     void writeTo(Encoder& encoder) const override;
     void resetFrom(Decoder& decoder) override;
 
     std::unique_ptr<std::string>    name_;
     std::unique_ptr<Expression>     arg_;
     bool                            distinct_{false};
     AggData*                        aggData_{nullptr};
+
+    // runtime cache for aggregate function lambda
+    folly::Optional<AggFunctionManager::AggFunction> aggFunc_;
```
Shylock-Hg, line 96: If the invalid function name is disabled in parser, I think you could remove the optional.

yixinglu (commented):
Lookup my comments in #407 , my suggestion is that never to check whether this function is nullable in `eval` function. It's in the hot path when evaluate expression. Please ensure the function existence when construct the aggr func call expression.

src/common/expression/AggregateExpression.h
czpmango: ACK.

czpmango:
> Lookup my comments in #407 , my suggestion is that never to check whether this function is nullable in `eval` function. It's in the hot path when evaluate expression. Please ensure the function existence when construct the aggr func call expression.

ACK.

jievince (approved):

src/common/expression/AggregateExpression.h
```diff
@@ -23,6 +23,12 @@ class AggregateExpression final : public Expression {
         arg_.reset(arg);
         name_.reset(name);
         distinct_ = distinct;
+        if (name_) {
+            auto aggFuncResult = AggFunctionManager::get(*name_);
+            if (aggFuncResult.ok()) {
+                aggFunc_ = std::move(aggFuncResult).value();
+            }
+        }
```
yixinglu, line 31: statici StatusOr<AggregateExpression*> create(name,xxx) {
  fn = AggFmgr.get(name);
  if (!fn) return Status::Error();
  return new AggregateFnExpr(fn);
}

czpmango: ACK.

czpmango: As we discussed offline, I tried to modify the interface above but got stuck by the memory management conflict of the parser, so the interface here is still maintained. Modify it if we unify the `Expression::create` interface later.

yixinglu (dismissed):

nevermore3 (dismissed):

yixinglu (approved):

Aiee (approved):

PR #473, opened 2021-03-29
insert vertex if not exists

nevermore3:
wait storage & graph..


src/common/interface/storage.thrift
```diff
@@ -406,8 +406,9 @@ struct AddVerticesRequest {
     //   in the NewVertex.NewTag.props
     3: map<common.TagID, list<binary>>
         (cpp.template = "std::unordered_map")   prop_names,
+    4: bool                                     if_not_exists,
     // If true, it equals an (up)sert operation.
-    4: bool                                     overwritable = true,
+    5: bool                                     overwritable = true,
```
panda-sheep, line 411: What is the difference between if_not_exists and overwritable?

nevermore3: when the (vertexID, tagID) already exists,   overwriteable do update and if_not_exists do nothing

panda-sheep: In my opinion, overwriteable is not currently used,
The effect of overwriteable should be the same as if_not_exists. WDYT?

nevermore3: Agree, maybe we can delete overwriteable, if_not_exist is easier to understand

panda-sheep: Another question
Here is to write records to the table, why only process the `tag `table, not the `edge `table?

As follows:
image

nevermore3: do it later

laura-ding (approved):

panda-sheep (approved):

yixinglu (approved):

PR #479, opened 2021-03-31
Add VidExpression

laura-ding:


src/common/expression/EdgeExpression.h
```diff
@@ -30,7 +36,7 @@ class EdgeExpression final : public Expression {
     }
 
     std::string toString() const override {
-        return "EDGE";
```
Shylock-Hg, line 33: Why don't return 'edge' directly?

src/common/expression/VertexExpression.h
```diff
@@ -30,7 +36,7 @@ class VertexExpression final : public Expression {
     }
 
     std::string toString() const override {
-        return "VERTEX";
```
Shylock-Hg, line 33: ditto

src/common/expression/EdgeExpression.h
```diff
@@ -46,6 +52,7 @@ class EdgeExpression final : public Expression {
 
 private:
     Value                                   result_;
+    std::unique_ptr<std::string>            outPutName_;
```
monadbobo, line 55: maybe "name_" was better.

src/common/expression/EdgeExpression.h
laura-ding: Because  the user can set the return name without aliai, just from the label

Shylock-Hg: In fact, better to change the return name by alias instead of change the expression.

PR #485, opened 2021-04-07
Send leader term in heart beat

liuyu85cn:
as title

related storage pr(this two have to merge together):
https://github.com/vesoft-inc/nebula-storage/pull/418


Shylock-Hg:
Why expose the internal info of raft?

liuyu85cn:
> Why expose the internal info of raft?

lang story, discussed offline at Beijing with Sherman. 
Actually this is for the BASE TOSS implement.

normally, we can use mvcc to prevent an old value overwrite a new one.
But, nebula now write mvcc info in value, which means when I want to check
which value is newer, I need to do a read.
Which is not acceptable which I do insert.

And we get a solution(use term) to skip most of the mvcc value check. 
That's why I have to expose term first.

critical27 (approved):

panda-sheep (approved):

monadbobo (approved):
LGTM.

PR #488, opened 2021-04-07
Reuse the date/time in datetime struct.

Shylock-Hg:
Will affect the client usage for the datetime structure change.

### DIFF
--- a/src/common/datatypes/Date.cpp
+++ b/src/common/datatypes/Date.cpp
@@ -107,8 +107,7 @@ std::string Time::toString() const {
 
 std::string DateTime::toString() const {
     // It's in current timezone already
-    return folly::stringPrintf(
-        "%d-%02d-%02dT%02d:%02d:%02d.%d", year, month, day, hour, minute, sec, microsec);
+    return folly::stringPrintf("%sT%s", d.toString().c_str(), t.toString().c_str());
 }
 
 }   // namespace nebula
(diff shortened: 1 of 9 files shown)


yixinglu (commented):
Is this changes compatible for data storage when next upgrading ?

Shylock-Hg:
> Is this changes compatible for data storage when next upgrading ?

Don't hurt the data format stored in disk.But will affect the graph interface (e.g. various clients ...)

panda-sheep (commented):
I think this pr maybe doing a wrong thing.

Because the time part of the datetime type and the time type are not the same thing.
First of all, their scope is different.
You can refer to the mysql database to have a look.

Shylock-Hg:
> I think this pr maybe doing a wrong thing.
> 
> Because the time part of the datetime type and the time type are not the same thing.
> First of all, their scope is different.
> You can refer to the mysql database to have a look.

Mysql require the more wide range value for represent duration of time, but the opencypher has specific type `Duration` for this. And the date/datetime/time range in opencypher is limited.

panda-sheep:
> Mysql require the more wide range value for represent duration of time, but the opencypher has specific type `Duration` for this. And the date/datetime/time range in opencypher is limited.

in opencypher, is the range of the time part of the datetime type the same as that of the time type?

Shylock-Hg:
> > Mysql require the more wide range value for represent duration of time, but the opencypher has specific type `Duration` for this. And the date/datetime/time range in opencypher is limited.
> 
> in opencypher, is the range of the time part of the datetime type the same as that of the time type?

Of course.

src/common/datatypes/Date.cpp
```diff
@@ -131,14 +130,14 @@ std::size_t hash<nebula::Time>::operator()(const nebula::Time& h) const noexcept
 }
 
 std::size_t hash<nebula::DateTime>::operator()(const nebula::DateTime& h) const noexcept {
-    size_t hv = folly::hash::fnv64_buf(reinterpret_cast<const void*>(&h.year), sizeof(h.year));
-    hv = folly::hash::fnv64_buf(reinterpret_cast<const void*>(&h.month), sizeof(h.month), hv);
-    hv = folly::hash::fnv64_buf(reinterpret_cast<const void*>(&h.day), sizeof(h.day), hv);
-    hv = folly::hash::fnv64_buf(reinterpret_cast<const void*>(&h.hour), sizeof(h.hour), hv);
-    hv = folly::hash::fnv64_buf(reinterpret_cast<const void*>(&h.minute), sizeof(h.minute), hv);
-    hv = folly::hash::fnv64_buf(reinterpret_cast<const void*>(&h.sec), sizeof(h.sec), hv);
+    size_t hv = folly::hash::fnv64_buf(reinterpret_cast<const void*>(&h.d.year), sizeof(h.d.year));
```
monadbobo, line 133: The size of size_t is not necessarily 64 bits.
`The bit width of std::size_t is not less than 16.`

Shylock-Hg: Fixed.

src/common/interface/common.thrift
```diff
@@ -69,13 +69,8 @@ struct Time {
 // !! Struct DateTime has a shadow data type defined in the Date.h
 // So any change here needs to be reflected to the shadow type there
 struct DateTime {
```
laura-ding, line 71: I think it is not necessary to modify, do not easily change the interface, otherwise it will be unfriendly to users.

Shylock-Hg: Yes, could add in next break version.

Shylock-Hg:
Stale

PR #489, opened 2021-04-08
[Simple] Add isArithmeticExpr() to expression

Aiee:
As title.

### DIFF
--- a/src/common/expression/Expression.h
+++ b/src/common/expression/Expression.h
@@ -146,6 +146,10 @@ public:
         return false;
     }
 
+    virtual bool isArithmeticExpr() const {
+        return false;
+    }
+
 protected:
     class Encoder final {
     public:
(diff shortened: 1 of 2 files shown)


src/common/expression/Expression.h
```diff
@@ -146,6 +146,10 @@ class Expression {
         return false;
     }
 
+    virtual bool isArithmeticExpr() const {
```
Shylock-Hg, line 149: You could check it by kind_

Aiee: There are multiple kinds of expressions that belong to arithmetic expressions, and every time when we need to check the expression type, we have to iterate all of them.

Shylock-Hg: It's more friendly to cpu cache.

Aiee: This function will only be called before the actual calculation of the expression and won't be involved in a hot spot.

yixinglu (approved):

nevermore3 (approved):

PR #491, opened 2021-04-08
fix substr crash

nevermore3:
refer to https://github.com/vesoft-inc/nebula-graph/issues/878

### DIFF
--- a/src/common/function/FunctionManager.cpp
+++ b/src/common/function/FunctionManager.cpp
@@ -1313,10 +1313,6 @@ FunctionManager::FunctionManager() {
         attr.isPure_ = true;
         attr.body_ = [](const auto &args) -> Value {
             auto argSize = args.size();
-            if (argSize < 2 || argSize > 3) {
-                LOG(ERROR) <<  "Unexpected arguments count " << args.size();
-                return Value::kNullBadData;
-            }
             if (args[0].isNull()) {
                 return Value::kNullValue;
             }
@@ -1325,11 +1321,18 @@ FunctionManager::FunctionManager() {
             }
             auto value = args[0].getStr();
             auto start = args[1].getInt();
-            auto length =  (args.size() == 2) ? value.size() - start : args[2].getInt();
+            auto length = 0;
+            if (argSize == 3) {
+                length = args[2].getInt();
+            } else {
+                length = static_cast<size_t>(start) >= value.size()
+                             ? 0
+                             : value.size() - static_cast<size_t>(start);
+            }
             if (start < 0 || length < 0) {
                 return Value::kNullBadData;
             }
-            if (static_cast<size_t>(start) > value.size() || length == 0) {
+            if (static_cast<size_t>(start) >= value.size() || length == 0) {
                 return std::string("");
             }
             return value.substr(start, length);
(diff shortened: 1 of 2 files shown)


src/common/function/FunctionManager.cpp
```diff
@@ -1313,32 +1313,26 @@ FunctionManager::FunctionManager() {
         attr.isPure_ = true;
         attr.body_ = [](const auto &args) -> Value {
             auto argSize = args.size();
-            if (argSize < 2 || argSize >3) {
+            if (argSize < 2 || argSize > 3) {
```
Shylock-Hg, line 1316: The find method will check the count of arguments.

src/common/function/FunctionManager.cpp
```diff
@@ -1313,32 +1313,26 @@ FunctionManager::FunctionManager() {
...
                 LOG(ERROR) <<  "Unexpected arguments count " << args.size();
                 return Value::kNullBadData;
             }
+            if (args[0].isNull()) {
```
laura-ding, line 1320: `empty` is the same. but the type Signature has made sure the input args[0] must string.

src/common/function/FunctionManager.cpp
nevermore3: fix

src/common/function/FunctionManager.cpp
nevermore3: type Signature only use in deduce expr's type , When it encounters null, it returns directly in validator
So in the execution stage, we need to re-judge

user cannot enter EMPTY .so we not process EMPTY

laura-ding: the query input can be empty, such as pipe

nevermore3: EMPTY will do later ,all function have this problem

laura-ding: Okay

src/common/function/test/FunctionManagerTest.cpp
```diff
@@ -150,6 +150,11 @@ TEST_F(FunctionManagerTest, testNull) {
     TEST_FUNCTION(toUpper, args_["nullvalue"], Value::kNullValue);
     TEST_FUNCTION(split, std::vector<Value>({Value::kNullValue, ","}), Value::kNullValue);
     TEST_FUNCTION(split, std::vector<Value>({"123,22", Value::kNullValue}), Value::kNullValue);
+    TEST_FUNCTION(substr, std::vector<Value>({Value::kNullValue, 1, 2}), Value::kNullValue);
+    TEST_FUNCTION(substr, std::vector<Value>({"hello", Value::kNullValue, 2}), Value::kNullBadType);
+    TEST_FUNCTION(substr, std::vector<Value>({"hello", 2, Value::kNullValue}), Value::kNullBadType);
+    TEST_FUNCTION(substr, std::vector<Value>({"hello", -1, 10}), Value::kNullBadData);
+    TEST_FUNCTION(substr, std::vector<Value>({"hello", 1, -2}), Value::kNullBadData);
```
laura-ding, line 157: add test when `start` large than string's size

nevermore3: alread exist in line 244

laura-ding (approved):

yixinglu (approved):

PR #504, opened 2021-04-15
Transfer ownership between objectpools

yixinglu:
Sometimes we need to transfer some object ownerships to another pool for releasing current pool early. e.g. in optimizer of graph, we can only save the used plan nodes in `QueryContext` when finishing optimization, other nodes could be destroyed with the release of OptGroups.

### DIFF
--- a/src/common/base/ObjectPool.h
+++ b/src/common/base/ObjectPool.h
@@ -8,7 +8,7 @@
 #define UTIL_OBJECTPOOL_H_
 
 #include <functional>
-#include <list>
+#include <unordered_map>
 
 #include <folly/SpinLock.h>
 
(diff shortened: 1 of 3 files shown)


src/common/base/ObjectPool.h
```diff
@@ -18,51 +18,52 @@ namespace nebula {
...
-    // Holder the ownership of the any object
-    class OwnershipHolder {
-    public:
-        template <typename T>
-        explicit OwnershipHolder(T *obj)
-            : obj_(obj), deleteFn_([](void *p) { delete reinterpret_cast<T *>(p); }) {}
-
-        ~OwnershipHolder() {
-            deleteFn_(obj_);
+    DeleteFn release(void *obj) {
```
Shylock-Hg, line 47: I think you could move the OwernshipHolder out directly.

yixinglu: It's slow to find the pointer in list.

Shylock-Hg: I don't say about the container(list or map). I means you could replace list by map, but still store the OwnershipHolder for the ownership management.

laura-ding (approved):

Aiee (approved):

monadbobo (approved):
LGTM.

PR #509, opened 2021-04-20
[opencypher compatibility] Support using exists with dynamic map

Aiee:
As title.

Dependency of https://github.com/vesoft-inc/nebula-graph/pull/973

### DIFF
--- a/src/common/expression/PredicateExpression.cpp
+++ b/src/common/expression/PredicateExpression.cpp
@@ -17,10 +17,18 @@ std::unordered_map<std::string, PredicateExpression::Type> PredicateExpression::
 };
 
 const Value& PredicateExpression::evalExists(ExpressionContext& ctx) {
-    DCHECK(collection_->kind() == Expression::Kind::kAttribute);
-    auto* attributeExpr = static_cast<AttributeExpression* >(collection_.get());
-    auto& container = attributeExpr->left()->eval(ctx);
-    auto& key = attributeExpr->right()->eval(ctx);
+    DCHECK(collection_->kind() == Expression::Kind::kAttribute ||
+           collection_->kind() == Expression::Kind::kSubscript);
+
+    BinaryExpression* containerExpr;
+    if (collection_->kind() == Expression::Kind::kAttribute) {
+        containerExpr = static_cast<AttributeExpression*>(collection_.get());
+    } else {
+        containerExpr = static_cast<SubscriptExpression*>(collection_.get());
+    }
+
+    auto& container = containerExpr->left()->eval(ctx);
+    auto& key = containerExpr->right()->eval(ctx);
     switch (container.type()) {
         case Value::Type::VERTEX: {
             result_ = container.getVertex().contains(key);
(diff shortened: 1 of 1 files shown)


src/common/expression/PredicateExpression.cpp
```diff
@@ -17,10 +17,18 @@ std::unordered_map<std::string, PredicateExpression::Type> PredicateExpression::
...
-    auto* attributeExpr = static_cast<AttributeExpression* >(collection_.get());
-    auto& container = attributeExpr->left()->eval(ctx);
-    auto& key = attributeExpr->right()->eval(ctx);
+    DCHECK(collection_->kind() == Expression::Kind::kAttribute ||
+           collection_->kind() == Expression::Kind::kSubscript);
+
+    BinaryExpression* containerExpr;
+    if (collection_->kind() == Expression::Kind::kAttribute) {
+        containerExpr = static_cast<AttributeExpression*>(collection_.get());
+    } else {
```
Shylock-Hg, line 26: else if ( == kSubscript)
else {
LOG(FATAL)
}

src/common/expression/PredicateExpression.cpp
```diff
@@ -17,10 +17,18 @@ std::unordered_map<std::string, PredicateExpression::Type> PredicateExpression::
 };
 
 const Value& PredicateExpression::evalExists(ExpressionContext& ctx) {
-    DCHECK(collection_->kind() == Expression::Kind::kAttribute);
-    auto* attributeExpr = static_cast<AttributeExpression* >(collection_.get());
-    auto& container = attributeExpr->left()->eval(ctx);
-    auto& key = attributeExpr->right()->eval(ctx);
+    DCHECK(collection_->kind() == Expression::Kind::kAttribute ||
```
Shylock-Hg, line 20: And this could remove.

nevermore3 (approved):

src/common/expression/PredicateExpression.cpp
Aiee: The expression type has been checked in the parser:
```
KW_EXISTS L_PAREN expression R_PAREN {
        if ($3->kind() != Expression::Kind::kLabelAttribute && $3->kind() != Expression::Kind::kAttribute &&
            $3->kind() != Expression::Kind::kSubscript) {
            throw nebula::GraphParser::syntax_error(@3, "The exists only accept LabelAttribe OR Attribute");
        }
        $$ = new PredicateExpression(new std::string("exists"), nullptr, $3, nullptr);
```
Is it still necessary to use the if-else here?

Shylock-Hg: There is if-else already.

Aiee: The `LOG(FATAL)` will never be printed because of the check from the parser tho.

czpmango: IMO, if-else block is not necessary if you only need to get its operands.

nevermore3 (approved):

yixinglu (approved):

jievince (approved):

PR #512, opened 2021-04-21
fix some relational expressions with Null/badNull

czpmango:
Depended by [https://github.com/vesoft-inc/nebula-graph/pull/972](https://github.com/vesoft-inc/nebula-graph/pull/972)

### DIFF
--- a/src/common/expression/RelationalExpression.cpp
+++ b/src/common/expression/RelationalExpression.cpp
@@ -39,9 +39,7 @@ const Value& RelationalExpression::eval(ExpressionContext& ctx) {
                 result_ = Value::kNullBadType;
             } else if ((!lhs.isNull() && !lhs.isStr()) || (!rhs.isNull() && !rhs.isStr())) {
                 result_ = Value::kNullBadType;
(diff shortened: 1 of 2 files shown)


src/common/expression/RelationalExpression.cpp
```diff
@@ -125,58 +125,82 @@ const Value& RelationalExpression::eval(ExpressionContext& ctx) {
             break;
         }
         case Kind::kContains: {
-            if (lhs.isStr() && rhs.isStr()) {
+            if (lhs.isBadNull() || rhs.isBadNull()) {
```
nevermore3, line 128: can the logic be simplified?
if (lhs.isNull() || rhs.isNull()) {
    result_ = Value::kNullValue;
} else if (lhs.isStr() && rhs.isStr()) {
    do something
} else {
   result_ = value::kNullBadType;
}

Aiee (approved):

CPWstatic (approved):

CPWstatic (approved):

PR #513, opened 2021-04-21
remove kvstore::ResultCode

panda-sheep:
Only For removing `kvstore::ResultCode`
https://github.com/vesoft-inc/nebula-graph/pull/1022 and https://github.com/vesoft-inc/nebula-storage/pull/448 reply on this pr.

### DIFF
--- a/src/common/interface/common.thrift
+++ b/src/common/interface/common.thrift
@@ -210,111 +210,173 @@ struct PartitionBackupInfo {
  */
 enum ErrorCode {
     // for common code
-    SUCCEEDED                   = 0,
-    E_DISCONNECTED              = -1,        // RPC Failure
-    E_FAIL_TO_CONNECT           = -2,
-    E_RPC_FAILURE               = -3,
-    E_LEADER_CHANGED            = -4,
-    E_INVALID_PARM              = -5,
-    E_REBUILD_INDEX_FAILED      = -6,
-    E_BACKUP_FAILED             = -7,        // Backup Failure
-    E_SPACE_NOT_FOUND           = -8,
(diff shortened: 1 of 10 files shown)


monadbobo (dismissed):
LGTM.

panda-sheep:
Replace this pr with https://github.com/vesoft-inc/nebula-common/pull/517，so close

src/common/interface/common.thrift
```diff
@@ -200,3 +200,121 @@ struct LogInfo {
...
 }
+
+
+/*
+ * ErrorCode for graphd, metad, storaged,raftd
+ * -1xxx for graphd
+ * -2xxx for metad
+ * -3xxx for storaged
+ */
+enum ErrorCode {
```
Shylock-Hg, line 211: We'd better value the number in hex format that we could easy check the type of error code by bit operatiion.

panda-sheep: Why need bit operations to check the type of error code?

panda-sheep:
Ready to review again.

src/common/interface/common.thrift
Shylock-Hg: To check where the error from

panda-sheep: There will be a special pr to unify the error code later.
It can be ignored here for now.

critical27 (approved):
Good job!

panda-sheep:
fix rebase master build failed

bright-starry-sky (approved):
LGTM

critical27 (approved):

PR #520, opened 2021-04-23
Concat & concat_ws 

nevermore3:
refer to https://discuss.nebula-graph.com.cn/t/topic/3181

### DIFF
--- a/src/common/function/FunctionManager.cpp
+++ b/src/common/function/FunctionManager.cpp
@@ -30,7 +30,6 @@ FunctionManager &FunctionManager::instance() {
 std::unordered_map<std::string, Value::Type> FunctionManager::variadicFunReturnType_ = {
     {"concat", Value::Type::STRING},
     {"concat_ws", Value::Type::STRING},
-    {"udf_is_in", Value::Type::BOOL},
     {"cos_similarity", Value::Type::FLOAT},
 };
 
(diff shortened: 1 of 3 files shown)


src/common/function/FunctionManager.cpp
```diff
@@ -27,6 +27,13 @@ FunctionManager &FunctionManager::instance() {
     return instance;
 }
 
+std::unordered_map<std::string, Value::Type> FunctionManager::variadicFunReturnType_ = {
+    {"concat", Value::Type::STRING},
+    {"concat_ws", Value::Type::STRING},
+    {"udf_is_in", Value::Type::BOOL},
```
monadbobo, line 33: Why recreate a map instead of continuing to use typeSignature_?

nevermore3: The number of parameters for these functions is uncertain，Can't list them all ，So The parameters are not processed in the verification phase and the results are returned directly, and the parameters are processed in the execution phase

Shylock-Hg: I think you could add a new field like `bool variadic_` in TypeSignature` to avoid store the signature in two map.

Shylock-Hg:
What's the difference between `concat` and `concat_ws`?

src/common/function/FunctionManager.cpp
nevermore3: It is better to distinguish, and the logic is clearer

nevermore3:
> What's the difference between `concat` and `concat_ws`?

yixinglu (approved):

yixinglu (approved):

PR #521, opened 2021-04-23
Cleanup std::string pointer usage in expression

yixinglu:
It's not necessary to hold a pointer of std::string in expression, we can copy the string contents and free the original memory in graph parser. I really don't want to create some new strings every time make a new expression. that's a too ugly practice.


czpmango (approved):

Shylock-Hg:
I think you could simply add a ctor like `Expr("","")`, allocate too much in parser is not efficient.

yixinglu:
> I think you could simply add a ctor like `Expr("","")`, allocate too much in parser is not efficient.

Not agree with you, We build expression trees in other modules much more times than in the parser.  It's more important in other modules to build the expression tree conveniently.

Aiee (approved):

PR #525, opened 2021-04-28
Add the subscript range expression.

Shylock-Hg:


src/common/expression/Expression.h
```diff
@@ -49,6 +49,7 @@ class Expression {
         kNotEndsWith,
 
         kSubscript,
+        kSubscriptRange,
```
yixinglu, line 52: This will crash the nebula upgrade since the expression'll be encoded into metad storage for default value expression.

append it to this enumeration type.

src/common/expression/SubscriptExpression.cpp
```diff
@@ -116,4 +118,159 @@ void SubscriptExpression::accept(ExprVisitor *visitor) {
...
+// For the positive range bound it start from begin,
+// for the negative range bound it start from end
+const Value& SubscriptRangeExpression::eval(ExpressionContext &ctx) {
+    const auto listValue = DCHECK_NOTNULL(list_)->eval(ctx);
+    if (!listValue.isList()) {
+        result_ = Value::kNullBadType;
+        return result_;
+    }
+    const auto list = listValue.getList();
+    int64_t lo, hi;
```
yixinglu, line 130: prefer upper than high

src/common/expression/SubscriptExpression.h
```diff
@@ -32,6 +33,89 @@ class SubscriptExpression final : public BinaryExpression {
...
+    void writeTo(Encoder& encoder) const override;
+
+    void resetFrom(Decoder& decoder) override;
+
+    // It's must list typed
+    std::unique_ptr<Expression> list_;
+
+    // range is [lo, hi), they could be any integer
+    std::unique_ptr<Expression> lo_;
+    std::unique_ptr<Expression> hi_;
```
yixinglu, line 113: ditto

src/common/expression/test/ExpressionTest.cpp
```diff
@@ -2207,6 +2207,87 @@ TEST_F(ExpressionTest, ListSubscript) {
...
+    }
+    // [0,1,2,3,4,5][-2..-3] => []
+    {
+        auto *lo = new ConstantExpression(-2);
+        auto *hi = new ConstantExpression(-3);
+        SubscriptRangeExpression expr(list->clone().release(), lo, hi);
+        auto value = Expression::eval(&expr, gExpCtxt);
+        EXPECT_TRUE(value.isList());
+        EXPECT_EQ(List(), value.getList());
+    }
```
yixinglu, line 2288: Need some test cases:

1. [0,1,2,3,4,5][2..-3]
2. [0,1,2,3,4,5][-2..3]

yixinglu (commented):
Good job !

src/common/expression/Expression.h
Shylock-Hg: Good catch. Fixed.

src/common/expression/SubscriptExpression.cpp
Shylock-Hg: These are used widely too.

jievince (approved):

yixinglu (approved):

PR #528, opened 2021-05-10
fix type

darionyaphet:


laura-ding (commented):
the nebula-graph and nebula-storage also need to modify

yixinglu (approved):

PR #531, opened 2021-05-14
Fix Integer overflow

Aiee:
As title.

```
(root@nebula) [nba]> yield 9223372036854775807 + 1
[ERROR (-8)]: (9223372036854775807+1) cannot be represented as an integer

(root@nebula) [nba]> yield 9223372036854775807/0
[ERROR (-8)]: / by zero
```
Required by https://github.com/vesoft-inc/nebula-graph/pull/1043

### DIFF
--- a/src/common/datatypes/Value.cpp
+++ b/src/common/datatypes/Value.cpp
@@ -1925,11 +1925,11 @@ Value operator+(const Value& lhs, const Value& rhs) {
                 case Value::Type::INT: {
                     auto lVal = lhs.getInt();
                     auto rVal = rhs.getInt();
-                    if ((lVal > 0 && rVal > INT64_MAX - lVal) ||
-                        (lVal < 0 && rVal < INT64_MIN - lVal)) {
+                    int sum;
+                    if (__builtin_add_overflow(lVal, rVal, &sum)) {
                         return Value::kNullOverflow;
                     }
-                    return lVal + rVal;
+                    return sum;
                 }
                 case Value::Type::FLOAT: {
                     return lhs.getInt() + rhs.getFloat();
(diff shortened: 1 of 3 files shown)


Shylock-Hg:
Do you test in ngql?

Aiee:
> Do you test in ngql?

Ngql related tests will be added in the graph PR.

src/common/datatypes/Value.cpp
```diff
@@ -1923,7 +1923,13 @@ Value operator+(const Value& lhs, const Value& rhs) {
         case Value::Type::INT: {
             switch (rhs.type()) {
                 case Value::Type::INT: {
-                    return lhs.getInt() + rhs.getInt();
+                    auto lVal = lhs.getInt();
+                    auto rVal = rhs.getInt();
+                    if ((lVal > 0 && rVal > INT64_MAX - lVal) ||
+                        (lVal < 0 && rVal < INT64_MIN - lVal)) {
```
monadbobo, line 1929: Perhaps it would be better if we used the compiler's built-in functions.

https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html
https://clang.llvm.org/docs/LanguageExtensions.html#checked-arithmetic-builtins

Aiee: I also checked these built-in functions, however, we don't know which compiler the user will use. I believe the implementation in this PR is sufficient to check overflow as well as being universal. 
WDYT?

monadbobo: These functions are supported by gcc after 5.x and by clang after 4.x. If you are not sure, then you can add detection code to cmake.

Aiee: Build-in functions have been used.

monadbobo (approved):
LGTM.

src/common/datatypes/Value.cpp
```diff
@@ -1923,7 +1923,13 @@ Value operator+(const Value& lhs, const Value& rhs) {
         case Value::Type::INT: {
             switch (rhs.type()) {
                 case Value::Type::INT: {
-                    return lhs.getInt() + rhs.getInt();
+                    auto lVal = lhs.getInt();
+                    auto rVal = rhs.getInt();
+                    int sum;
```
yixinglu, line 1928: int64_t

src/common/datatypes/Value.cpp
```diff
@@ -2171,7 +2177,13 @@ Value operator-(const Value& lhs, const Value& rhs) {
         case Value::Type::INT: {
             switch (rhs.type()) {
                 case Value::Type::INT: {
-                    return lhs.getInt() - rhs.getInt();
+                    auto lVal = lhs.getInt();
+                    auto rVal = rhs.getInt();
+                    int res;
```
yixinglu, line 2182: ditto

yixinglu (commented):
pls check all integer usage, change its type to `int64_t`

yixinglu (approved):

monadbobo (approved):
LGTM.

dutor (approved):
Bravo!

PR #534, opened 2021-05-18
Fix bug: emplace only do insertion when key does not exists.

guojun85:
Steps to reproduce this issue:
1. create a index:
(root@nebula) [basketballplayer]> create tag index name on player(name(20));
Execution succeeded (time spent 2989/3376 us)

2. drop the index:
(root@nebula) [basketballplayer]> drop tag index name
Execution succeeded (time spent 2750/3179 us)

3. create the index again:
(root@nebula) [basketballplayer]> create tag index name on player(name(20));
Execution succeeded (time spent 2989/3376 us)

4. drop the index again will return fails:
(root@nebula) [basketballplayer]> drop tag index name;
[ERROR (-12)]: SemanticError: index name was not found

the root cause is that  edgeNameIndexMap_.emplace(std::move(pair), indexID) does not update edgeNameIndexMap_ as old key exists.

### DIFF
--- a/src/common/clients/meta/MetaClient.cpp
+++ b/src/common/clients/meta/MetaClient.cpp
@@ -440,7 +440,7 @@ bool MetaClient::loadIndexes(GraphSpaceID spaceId,
         auto indexName = tagIndex.get_index_name();
         auto indexID = tagIndex.get_index_id();
         std::pair<GraphSpaceID, std::string> pair(spaceId, indexName);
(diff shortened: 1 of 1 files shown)


panda-sheep (approved):
Good catch

Shylock-Hg (approved):

monadbobo (approved):
LGTM.

PR #538, opened 2021-05-20
[Bug fix] fix src() and dst() for reversed edge

Aiee:
As title.

### DIFF
--- a/src/common/function/FunctionManager.cpp
+++ b/src/common/function/FunctionManager.cpp
@@ -1748,7 +1748,8 @@ FunctionManager::FunctionManager() {
                     return Value::kNullValue;
                 }
                 case Value::Type::EDGE: {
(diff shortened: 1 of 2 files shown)


src/common/function/FunctionManager.cpp
```diff
@@ -1748,7 +1748,8 @@ FunctionManager::FunctionManager() {
                     return Value::kNullValue;
                 }
                 case Value::Type::EDGE: {
-                    return args[0].getEdge().src;
+                    auto edge = args[0].getEdge();
```
yixinglu, line 1751: const auto &

Aiee: Fixed.

nevermore3 (approved):

yixinglu (approved):

PR #541, opened 2021-05-24
Add memory info to get system memory usage

yixinglu:
refer to vesoft-inc/nebula-graph#1052

### DIFF
--- a/src/common/base/Memory.cpp
+++ b/src/common/base/Memory.cpp
@@ -12,12 +12,22 @@
 
 namespace nebula {
 
+StatusOr<std::unique_ptr<MemInfo>> MemInfo::make() {
+    std::unique_ptr<MemInfo> mem(new MemInfo);
+    NG_RETURN_IF_ERROR(mem->init());
+    return mem;
+}
+
 MemInfo::MemInfo() noexcept {
     info_ = std::make_unique<struct sysinfo>();
+}
+
+Status MemInfo::init() {
     if (sysinfo(info_.get()) == -1) {
         auto err = errno;
-        LOG(ERROR) << "Fail to call sysinfo to get memory info, errno: " << err;
+        return Status::Error("Fail to call sysinfo to get memory info, errno: %d", err);
     }
+    return Status::OK();
 }
 
 }   // namespace nebula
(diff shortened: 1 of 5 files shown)


src/common/base/Memory.h
```diff
@@ -0,0 +1,46 @@
...
+#include <cstdint>
+
+#include "common/cpp/helpers.h"
+
+namespace nebula {
+
+class MemInfo final : protected cpp::NonCopyable, protected cpp::NonMovable {
+public:
+    MemInfo() noexcept {
+        sysinfo(&info_);
```
critical27, line 21: Could consider to check the return value here.

src/common/base/Memory.h
```diff
@@ -0,0 +1,46 @@
...
+
+    uint64_t freeInKB() const {
+        return (info_.freeram * info_.mem_unit) >> 10;
+    }
+
+    uint64_t usedInKB() const {
+        return totalInKB() - freeInKB();
+    }
+
+    bool hitsHighWatermark(float ratio = 0.9f) const {
```
critical27, line 36: Better set the same default value with `system_memory_high_watermark_ratio`.

src/common/base/Memory.h
yixinglu: now Log error when return value is -1.

src/common/base/Memory.h
yixinglu: reset to 0.8f

bright-starry-sky (approved):
LGTM

src/common/base/Memory.cpp
```diff
@@ -0,0 +1,33 @@
...
+    NG_RETURN_IF_ERROR(mem->init());
+    return mem;
+}
+
+MemInfo::MemInfo() noexcept {
+    info_ = std::make_unique<struct sysinfo>();
+}
+
+Status MemInfo::init() {
+    if (sysinfo(info_.get()) == -1) {
```
bright-starry-sky, line 26: What's about `if (sysinfo(info_.get()) != 0)` ?

dutor (approved):
LGTM.

PR #545, opened 2021-05-26
Add the hold ptr type to mark the ownership.

Shylock-Hg:

PR #546, opened 2021-05-27
Replace unique_ptr with raw pointer in expression implementation

Aiee:
1. Replace all unique_ptr(s) with raw pointers in expression implementation and manage object lifetime using ObjectPool.
2. Disable default expression constructors to preventing mixing raw pointer and unique_ptr.

TODOs:
- [x]  Replace expression constructor with static functions, specifying expression type, and add expression into ObjectPool during the construction.
- [x]  Split expression UTs to separate files
- [x]  Modify other related common UTs
- [x]  Encode/Decode test


Shylock-Hg:
The raw pointer is too implicit for the lifetime management. You could force everyone remember all object in expression is managed by objpool, but when these object referenced by other or passed from others it's more confuse.

src/common/expression/Expression.h
```diff
@@ -138,13 +141,17 @@ class Expression {
     virtual void accept(ExprVisitor* visitor) = 0;
 
     // Deep copy
-    virtual std::unique_ptr<Expression> clone() const = 0;
+    virtual Expression* clone() const = 0;
```
czpmango, line 144: Maybe need `virtual Expression* clone(ObjPool* pool) const = 0`, same for derived classes.

src/common/expression/ArithmeticExpression.h
```diff
@@ -13,26 +13,56 @@ namespace nebula {
...
 public:
-    ArithmeticExpression(Kind kind,
-                         Expression* lhs = nullptr,
-                         Expression* rhs = nullptr)
-        : BinaryExpression(kind, lhs, rhs) {}
+    static ArithmeticExpression* makeAdd(ObjectPool* pool = nullptr,
+                                         Expression* lhs = nullptr,
+                                         Expression* rhs = nullptr) {
+        DCHECK(!!pool);
+        return pool->add(new ArithmeticExpression(pool, Expression::Kind::kAdd, lhs, rhs));
```
czpmango, line 20: if `DCHECK`, You should remove the default parameters of the `pool`.

czpmango: Add `DCHECK_EQ(pool, lhs->getPool())` to ensure consisten ObjectPool, same for other classes.

src/common/expression/Expression.h
```diff
@@ -138,13 +141,17 @@ class Expression {
...
     std::string encode() const;
 
     static std::string encode(const Expression& exp);
 
-    static std::unique_ptr<Expression> decode(folly::StringPiece encoded);
+    static Expression* decode(ObjectPool* pool, folly::StringPiece encoded);
+
+    ObjectPool* getObjPool() const {
+        return pool_;
+    }
```
czpmango, line 154: It's better to delete the setter for ObjectPool explicitly to avoid misuse.   ObjectPool must be specified at construct or clone interface.

Aiee:
> The raw pointer is too implicit for the lifetime management. You could force everyone remember all object in expression is managed by objpool, but when these object referenced by other or passed from others it's more confuse.

After off-line discussion, this scenario could be solved by changing all expression-related members from `unique_ptr`s to raw pointers.

src/common/expression/Expression.h
Aiee: Currently, all expressions will be managed by the object pool in query context, and to reduce confusion, I believe there is no need to provide this interface.

src/common/expression/ArithmeticExpression.h
Aiee: Default parameters have been removed. And for the reason stated in the conversation below, `DCHECK_EQ(pool, lhs->getPool())` is unnecessary.

src/common/base/ObjectPool.h
```diff
@@ -29,6 +32,9 @@ class ObjectPool final : private cpp::NonCopyable, private cpp::NonMovable {
 
     template <typename T>
     T *add(T *obj) {
+        if constexpr (std::is_same_v<T, Expression>) {
```
Shylock-Hg, line 35: I don't think should print it.

Aiee: This makes it easier to detect the double-free problem when adding the same expression multiple times to the pool

Shylock-Hg: In fact, we could disable the double adding by type check. See the pr #545 , this separate the raw pointer and the object hold in pool.

CPWstatic (approved):

CPWstatic (approved):

src/common/base/ObjectPool.h
yixinglu: this check is useless for inherited expression. please remove it.

yixinglu (commented):
ltgm

yixinglu (approved):

PR #550, opened 2021-05-31
count buffer/cache into MemInfo

critical27:
`Total = Used + Free + Buffer/Cache`

Maybe need to take buffer/cache into consideration, here a screenshot when I do some test.

image

Graph will keep logging something like `Used memory(211303932KB) hits the high watermark(0.800000) of total system memory(263856036KB).`.

PR is still in verification.

### DIFF
--- a/src/common/base/Memory.h
+++ b/src/common/base/Memory.h
@@ -29,8 +29,12 @@ public:
         return (info_->freeram * info_->mem_unit) >> 10;
     }
 
+    uint64_t bufferInKB() const {
+        return (info_->bufferram * info_->mem_unit) >> 10;
+    }
+
     uint64_t usedInKB() const {
-        return totalInKB() - freeInKB();
+        return totalInKB() - freeInKB() - bufferInKB();
     }
 
     bool hitsHighWatermark(float ratio = 0.8f) const {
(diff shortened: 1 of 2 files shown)


bright-starry-sky (approved):
LGTM

panda-sheep (approved):
Good!

yixinglu:
We should count the physical memory usage, not include buffer. Here `freeInKB()` returns the same result as the free column of `free -h`.

More information refer to https://stackoverflow.com/questions/63166/how-to-determine-cpu-and-memory-consumption-from-inside-a-process

critical27:
> We should count the physical memory usage, not include buffer. Here `freeInKB()` returns the same result as the free column of `free -h`.
> 
> More information refer to https://stackoverflow.com/questions/63166/how-to-determine-cpu-and-memory-consumption-from-inside-a-process

Could have discuss later. Some known issues when I write data by importer:
1. Failed to write data once pressure if enough
2. Some interface could bypass the check later, `show hosts` will fail as well if high watermark is reached.

If take buffer into consideration, it will solve these problems.

yixinglu:
> > We should count the physical memory usage, not include buffer. Here `freeInKB()` returns the same result as the free column of `free -h`.
> > More information refer to https://stackoverflow.com/questions/63166/how-to-determine-cpu-and-memory-consumption-from-inside-a-process
> 
> Could have discuss later. Some known issues when I write data by importer:
> 
> 1. Failed to write data once pressure if enough
> 2. Some interface could bypass the check later, `show hosts` will fail as well if high watermark is reached.
> 
> If take buffer into consideration, it will solve these problems.

Good point for question 1. Now you can update this config to `1.0` before importing and restore the config when finishing data loading.

In query engine, we have ignored the checks for most executors except query related executors and if `show hosts` still report this error, that's a bug.

czpmango (approved):

PR #557, opened 2021-06-03
Workaround to fix clang build on aarch64

dutor:
Build fails in arm64 box with clang. It involves folly, clang, gcc and arm neon. So it's complicated. This patch is kind of a workaround.


yixinglu (approved):

Shylock-Hg (approved):

dutor:
Thank you guys!

PR #558, opened 2021-06-07
add process memory info stats

czpmango:
Depends by [https://github.com/vesoft-inc/nebula-graph/pull/1116](https://github.com/vesoft-inc/nebula-graph/pull/1116)


critical27 (commented):
I have do some research previously in https://github.com/critical27/nebula-common/tree/mem_tracker, you could take a look if you are interested. I didn't do any test except a unit test.

This feature need to be verified carefully.

PR #559, opened 2021-06-07
fix gflags conversion

czpmango:


laura-ding (approved):

laura-ding:
Take a look at the compilation failure problem

PR #563, opened 2021-06-15
Get tag by property.

Shylock-Hg:


src/common/meta/SchemaManager.h
```diff
@@ -10,6 +10,7 @@
 #include "common/base/Base.h"
 #include "common/base/StatusOr.h"
 #include <folly/RWSpinLock.h>
+#include <vector>
```
panda-sheep, line 13: Vector.h has been included in` #include "common/base/Base.h"`

src/common/meta/ServerBasedSchemaManager.cpp
```diff
@@ -4,6 +4,7 @@
  * attached with Common Clause Condition 1.0, found in the LICENSES directory.
  */
 
+#include <vector>
```
panda-sheep, line 7: ditto

src/common/meta/SchemaManager.h
Shylock-Hg: Fixed.

Aiee (approved):

PR #568, opened 2021-06-23
Avoid rehash for duplicate path check functions

yixinglu:
as title.

### DIFF
--- a/src/common/plugin/fulltext/FTGraphAdapter.h
+++ b/src/common/plugin/fulltext/FTGraphAdapter.h
@@ -8,6 +8,7 @@
 #define COMMON_PLUGIN_FULLTEXT_G_ADAPTER_H_
 
 #include "common/base/Base.h"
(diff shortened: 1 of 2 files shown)


Aiee (approved):

Shylock-Hg (approved):

nevermore3 (approved):

src/common/plugin/fulltext/FTGraphAdapter.h
```diff
@@ -8,6 +8,7 @@
 #define COMMON_PLUGIN_FULLTEXT_G_ADAPTER_H_
 
 #include "common/base/Base.h"
+#include "common/base/StatusOr.h"
```
Shylock-Hg, line 11: Why change this?

yixinglu: Used in following return type but miss to include it.

PR #573, opened 2021-06-28
Modify require cmake version

laura-ding:
close https://github.com/vesoft-inc/nebula-graph/issues/1181


yixinglu (approved):
nebula-graph also needs to upgrade the cmake version.

Aiee (approved):

PR #576, opened 2021-07-02
Add time out error code.

Shylock-Hg:


Aiee (approved):

src/common/graph/Response.h
```diff
@@ -48,6 +48,7 @@
     X(E_REBUILD_INDEX_FAILED            , -28)  \
     X(E_INVALID_PASSWORD                , -29)  \
     X(E_FAILED_GET_ABS_PATH             , -30)  \
+    X(E_RPC_TIMEOUT                     , -31)  \
```
laura-ding, line 51: add comment is the error code just for cpp client

Shylock-Hg: This could used by others too.

Aiee (approved):

Aiee:
Is it necessary to re-generate thrift files for clients?

Shylock-Hg:
> Is it necessary to re-generate thrift files for clients?

Don't force.

PR #579, opened 2021-07-07
Fix operator+ for string and float

Aiee:
As title.
The result should not present the float value as a fixed-digits value.

Close https://github.com/vesoft-inc/nebula-graph/issues/1214

### DIFF
--- a/src/common/datatypes/Value.cpp
+++ b/src/common/datatypes/Value.cpp
@@ -1966,9 +1966,8 @@ Value operator+(const Value& lhs, const Value& rhs) {
                     return lhs.getFloat() + rhs.getFloat();
                 }
                 case Value::Type::STRING: {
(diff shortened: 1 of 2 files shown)


src/common/datatypes/Value.cpp
```diff
@@ -1966,9 +1966,8 @@ Value operator+(const Value& lhs, const Value& rhs) {
                     return lhs.getFloat() + rhs.getFloat();
                 }
                 case Value::Type::STRING: {
-                    return folly::stringPrintf("%lf%s",
-                                               lhs.getFloat(),
-                                               rhs.getStr().c_str());
```
jievince, line 1971: remove folly::stringPrintf

jievince (approved):

HarrisChu (approved):

PR #585, opened 2021-07-14
Get var/input property by index.

Shylock-Hg:


src/common/expression/PropertyExpression.cpp
```diff
@@ -71,15 +71,29 @@ void TagPropertyExpression::accept(ExprVisitor* visitor) {
...
 const Value& VariablePropertyExpression::eval(ExpressionContext& ctx) {
-    return ctx.getVarProp(sym_, prop_);
+    if (!propIndex_.hasValue()) {
+        auto indexResult = ctx.getVarPropColIndex(sym_, prop_);
+        if (!indexResult.ok()) {
+            return Value::kEmpty;
+        }
+        propIndex_ = indexResult.value();
+    }
+    return ctx.getColumn(propIndex_.value());
```
yixinglu, line 96: I don't think it's a right way to use optional to improve the performance. you can give a setter function for prop index, and use the index in eval function directly to avoid check whether this value is valid each time.

Shylock-Hg: But when to set it ? Usually we just call Expression::eval and don't care the Input/Var Property expression.

PR #588, opened 2021-07-20
Adapte to kylin and mips64el

laura-ding:


src/common/base/ObjectPool.h
```diff
@@ -14,6 +14,7 @@
 #include <folly/SpinLock.h>
 
 #include "common/cpp/helpers.h"
+#include "common/base/Base.h"
```
yixinglu, line 17: cleanup this included header

laura-ding: The VLOG need it, but i think use `#include "common/base/Logging.h"` is better

yixinglu (approved):

darionyaphet (commented):
It's amazing

darionyaphet (approved):

PR #603, opened 2021-08-04
Delete tag

critical27:
As title

### DIFF
--- a/src/common/interface/storage.thrift
+++ b/src/common/interface/storage.thrift
@@ -363,13 +363,25 @@ struct DeleteVerticesRequest {
         (cpp.template = "std::unordered_map")           parts,
 }
 
-
 struct DeleteEdgesRequest {
     1: common.GraphSpaceID                      space_id,
     // partId => edgeKeys
     2: map<common.PartitionID, list<EdgeKey>>
         (cpp.template = "std::unordered_map")   parts,
 }
+
+struct DelTags {
+    1: common.Value id,
+    2: list<common.TagID> tags,
+}
+
+struct DeleteTagsRequest {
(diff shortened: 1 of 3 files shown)


Shylock-Hg:
Wether delete last tag equals to delete vertex?

critical27:
> Wether delete last tag equals to delete vertex?

For now the answer is no. This is a good point, I can't come up with a good way to do it at this time. Even if we have transactions, how to convert delete last tag to delete vertex could be tricky.

bright-starry-sky (approved):
LGTM.

src/common/interface/storage.thrift
```diff
@@ -363,13 +363,25 @@ struct DeleteVerticesRequest {
...
 struct DeleteEdgesRequest {
     1: common.GraphSpaceID                      space_id,
     // partId => edgeKeys
     2: map<common.PartitionID, list<EdgeKey>>
         (cpp.template = "std::unordered_map")   parts,
 }
+
+struct DelTags {
+    1: common.Value id,
+    2: list<common.TagID> tags,
```
panda-sheep, line 375: When deleting all tags of a vertex, what are the `tags `here?  Is it empty?

panda-sheep (commented):
good job.
There is a small question， see comments

src/common/interface/storage.thrift
panda-sheep: Yeah, I saw you did it on the graph layer.

critical27: Yeah

panda-sheep (approved):

PR #604, opened 2021-08-04
add storage profile detail

cangfengzhs:


src/common/interface/storage.thrift
```diff
@@ -35,6 +35,7 @@ struct ResponseCommon {
     1: required list<PartitionResult>   failed_parts,
     // Query latency from storage service
     2: required i32                     latency_in_us,
+    3: optional map<string,i32>         latency_detail_us,
```
critical27, line 38: Better unify with code in storage, which uses int64_t

cangfengzhs: "latency_in_us" is int32. Do they have to be the same?Ignore "latency_in_us" or change "latency_in_us" to int64?

critical27: Both int32_t and int64_t is ok, just unify code in thrift and storage will be fine.

critical27 (approved):

src/common/interface/storage.thrift
cangfengzhs: Ok, profile_detail_ map in storage has been changed.

cangfengzhs:
Has been merged into nebula repo.

